PR #10, opened 2019-04-23
Remote config and permissions

rlizzo:
PR to get the remote server in a more user-friendly form. 

Additions:
- config file for config of hangar server parameters. On startup, the client retrieves any necessary configuration parameters from the server itself, and uses those specified to establish the main channel connection. 
- Ability to make a remote server `clone` or `fetch` only for any non-authenticated users (authentication is handeld through client and server call interceptors.) THIS IS NOT MEANT TO BE A SECURE IMPLEMENTATION, and should not be regarded as such. 

Bugfixes:
- fixed issue where local commits consisting only of metadata additions may fail to push to a remote repo.
- Progress bars display more relevant information and now properly recycle screen real-estate.

Lots of minor improvements and optimizations along the way.


rlizzo:
Merging with failing tests due to severe testing config issues - which are being worked out separately from this PR. No changes here will affect core (local) hangar usage or implementation.

PR #14, opened 2019-04-29
Initial Testcases

hhsecond:


rlizzo:
@hhsecond let me know when this is ready for a code review. thanks for all the work!

rlizzo:
hey @hhsecond, 

As you'll see I pushed some changes on my end to update the tests and the test-runners. The travis-ci build is now passing on all tests (the appveyor config still needs some work, i'm not really sure what this issue with windows is right now). 

In general, really nice job! Definitely on our way to having decent test coverage here. I'll send in some more changes/comments when I have the next free minute, but this is a great start!

rlizzo:
see also #20 which we will want to merge and rebase this branch on top of.

src/hangar/config_hangar.yml
```diff
@@ -5,7 +5,7 @@ hangar:
   __version__: '0.0.0'
 
   lmdb:
-    map_size: 10_000_000_000
+    map_size: 50_000_000
```
rlizzo: So by reducing this to 50MB, we were able to get the windows test to start working (and failing for real reasons, not "out of disk space") apparently there's a large difference in how mmapped files are dealt with between unix and windows...

This needs to be reverted back before we merge. but just leaving it as a blocking issue here until we find a solution.

rlizzo:
I'm going to merge this so that focused test development can occur in isolated PRs.

Issue #12, opened 2019-04-26
Exception handling


hhsecond:
The way we are handling the exception is not user-friendly and has to be changed. This issue is to discuss it further. After thinking about it for some time, I think it's better to raise an exception rather than returning the exception object, as Rick suggested.  Here are the cases I have found out so far where we are not raising the exception instead of pushing it to stderr/stdout. I'll add more as I find

- Trying to init the already initialized dataset
- Trying to initiate dataset with more than 31 dimension
- Accessing dset['key_doesnt_exist']
- dset[1] = somearray (should raise the exception about integer being used as a key)
- wrong argument order to dset.add()
- If name is not passed to dset.add() for samples_are_named
- adding different dtypes and shapes than used while init
- committing without any changes should return something if not raising an exception

hhsecond:
Hey @rlizzo,  Have you got a chance to run through this? As far as I could think, raising an exception in all the cases (even for minor issues) is the best we could do unless we don't want to put the whole stack trace onto user's CLI (we could still use exception hooks to suppress those messages if we want). Or do you have some other reasons to use the bool return instead of raising exceptions that I might be missing?

rlizzo:
Hey @hhsecond, sorry I needed a few days to really think about this. 

I think you're completely right. We should be just raising exceptions for this. I'd like to stick to the Python standard library exception classes for now. As far as the stack traces go, they might print alot to the screen, but I think it's OK for now (if anything it'll help us debug any user bug reports); getting pretteir tracebacks is a problem for another time. If we just log to error to the console before raising the exception, it should be OK for now - we just need to have fairly informative error messages. 

I'm putting together a PR for this right now, should be merged before I go to bed tonight. sorry if this was holding you up at all!

hhsecond:
Hey @rlizzo, no worries and thanks a lot for taking this for me. I even had started this in a separate branch but then thought of discussing with you about the design. I am keeping this issue opened at least until I finish both local and remote testing. I'll open more PRs if required in the future given that now I know what should be the way forward for exceptions.

rlizzo:
ok. all these changes are merged in #16

rlizzo:
@hhsecond i'll leave it to you to close this issue when you feel it is appropriate. It's good for me right now.

rlizzo:
Closing due to inactivity

PR #15, opened 2019-04-29
[WIP]Minor fixes

hhsecond:
Keeping it open so that everyone is aware of these small fixes (until #14 goes live)

PR #16, opened 2019-04-30
Exception handling

rlizzo:
This PR addresses issues in #12. 

@hhsecond I'm going to merge this once you confirm that this covers the cases you need.

### DIFF
--- a/src/hangar/dataset.py
+++ b/src/hangar/dataset.py
@@ -1,9 +1,11 @@
 import hashlib
 import os
-from uuid import getnode
+from typing import Optional
 from uuid import uuid1
+import logging
 
 import numpy as np
+import lmdb
 
 from . import config
 from .context import TxnRegister
@@ -11,7 +13,7 @@ from .hdf5_store import FileHandles
 from .records import parsing
 from .records.queries import RecordQuery
 
-uuid_node = getnode()
+logger = logging.getLogger(__name__)
 
 
 class DatasetDataReader(object):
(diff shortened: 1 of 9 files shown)


src/hangar/dataset.py
```diff
@@ -670,68 +755,109 @@ def init_dataset(self, name: str, shape=None, dtype=None, prototype=None, *,
...
+            If rank of maximum tensor shape > 31.
         '''
-        if not (name.isalnum or name.isascii):
-            print(f'dataset name provided: {name} is not allowed. Must only contain '
-                  f'alpha-numeric ascii characters with no whitespace characters.')
-            return False
 
         # ------------- Checks for argument validity --------------------------
 
+        if not (name.isalnum or name.isascii):
```
hhsecond: - isalnum is a function and needs to be called
- isascii is a new feature came in 3.7 and won't work with 3.6

I have a half cooked PR with this fix and some others at #15

src/hangar/dataset.py
```diff
@@ -257,70 +267,116 @@ def __delitem__(self, key):
...
-        if tmpconman:
-            self.__enter__()
+        try:
+            if not isinstance(data, np.ndarray):
+                msg = f'HANGAR TYPE ERROR:: `data` argument type: {type(data)} != `np.ndarray`'
+                raise TypeError(msg)
+
+            if self._samples_are_named:
+                if not (isinstance(name, str)):
+                    msg = f'HANGAR VALUE ERROR:: no `name` argument provided while '\
```
hhsecond: IMO we should indicate the key should be string? Something like `HANGAR VALUE ERROR:: no name argument provided ... or provided value is not string` perhaps? Because this error is super confusing for somebody who makes an operation like 
```python
dset[1] = array
```

src/hangar/dataset.py
rlizzo: Great. i've fixed them here and will merge these so they are one logical set of additions.

src/hangar/dataset.py
rlizzo: Fixed!

rlizzo:
Merging this now. closes #16

Issue #5, opened 2019-04-20
Fetched data is not same as saved data


hhsecond:
In the same python session, if we initiate dataset twice and then save the data with same shape and type as the first instance, we won't get the data that we saved in the second instance. Instead, we'll get the data from first instance. Here is the script to reproduce.

```python
import numpy as np
from hangar.repository import Repository

test_hangar_path1 = '/home/hhsecond/mypro/test_hangar_dir_1'
arr1 = np.random.random((5, 7))
repo1 = Repository(path=test_hangar_path1)
repo1.init(user_name='tester1', user_email='foo@test1.bar', remove_old=True)
co = repo1.checkout(write=True)
dset1 = co.datasets.init_dataset('dset1', prototype=arr1)
dset1['arr1'] = arr1
co.commit()
co.close()


test_hangar_path2 = '/home/hhsecond/mypro/test_hangar_dir_2'
arr2 = np.random.random((5, 7))
repo2 = Repository(path=test_hangar_path2)
repo2.init(user_name='tester2', user_email='foo@test2.bar', remove_old=True)
co = repo2.checkout(write=True)
dset2 = co.datasets.init_dataset('dset2', prototype=arr2)
dset2['arr2'] = arr2
co.commit()
co.close()


co = repo2.checkout()
print(arr2[1], co.datasets['dset2']['arr2'][1])
print((arr2 == co.datasets['dset2']['arr2']).all())
```

rlizzo:
This is related to #4

Issue #4, opened 2019-04-20
Can't initiate two repository with different paths in the same python session


hhsecond:
I found an issue with initiating the repository again and again in the same python session even if the repository path is different. Because of the singleton nature of `Environment` class, regardless of the path, we pass for a repository, all the following repositories will have the path of the first repository (Is this something we desire to have?)

The below script can reproduce the issue. The last print statement must print `True`
```python
import numpy as np
from hangar.repository import Repository

test_hangar_path1 = '/home/hhsecond/mypro/test_hangar_dir_1'
arr1 = np.random.random((5, 7))
repo1 = Repository(path=test_hangar_path1)
repo1.init(user_name='tester1', user_email='foo@test1.bar', remove_old=True)
co = repo1.checkout(write=True)
dset1 = co.datasets.init_dataset('dset1', prototype=arr1)
dset1['arr1'] = arr1
co.commit()
co.close()


test_hangar_path2 = '/home/hhsecond/mypro/test_hangar_dir_2'
arr2 = np.random.random((5, 7))
repo2 = Repository(path=test_hangar_path2)
repo2.init(user_name='tester2', user_email='foo@test2.bar', remove_old=True)
co = repo2.checkout(write=True)
dset2 = co.datasets.init_dataset('dset2', prototype=arr2)
dset2['arr2'] = arr2
co.commit()
co.close()


print(test_hangar_path2, repo2._repo_path)
print(test_hangar_path2 == repo2._repo_path)
```

rlizzo:
So I think for right now it would be best to not allow this usage pattern. It's something I would like to fix, but I don't think it's particularly high on the priority list right now. 

To really solve this problem we should get rid of the Singleton's scattered around the codebase. They could probably be removed in the hdf5FileHandler and TransactionRegister with only mild headaches if we were to tie their lifespan to the Repository instance. The real issue is for the Environments class which sets up lmdb envs. Lmdb dbs absolutely need to only be opened once per process, or really bad things happened (the behavior is rather well documented, but i learned just how bad it gets the hard way...) Singleton's were my solution to that particular headache, but if you have another idea I'd love to hear it!

In lieu of a redesign at the Environment level, I think we should just raise an error if a second repo instance is created in a particular process. That would be pretty easy to implement.

Thoughts?

rlizzo:
So i may have to eat my words here a bit with #20. It actually seems to be a decent solution to this issue... though we still . will want to kill the singletons hanging around the codebase sometime in the future.

PR #20, opened 2019-04-30
Allow single process access to multiple repos.

rlizzo:
Fix behavior which prevented a single python process from working on multiple repositories. 

## Description

These changes modify the behavior of the Environment and hdf5 FileHandler singleton classes to instantiate singleton instances based on the path of the checked out repository. This seems to fix all outstanding bugs related to these odd behaviors, but is untested as of yet.

The TransactionRegister singleton did not need to be updated because it the reference counting scheme is isolated for each lmdb environment which is passed to it. As different repository instances in the same process will now have unique lmdb environment handles, it does not require any modifications to it's current behavior.

## Motivation and Context
Addresses:
- https://github.com/tensorwerk/hangar-py/issues/4
- https://github.com/tensorwerk/hangar-py/issues/5

## How Has This Been Tested?
- local, and at head of #14 

## Types of changes

## Checklist:

ping @hhsecond for sanity check.

### DIFF
--- a/src/hangar/context.py
+++ b/src/hangar/context.py
@@ -175,9 +175,10 @@ from .records import commiting, heads
 class EnvironmentsSingleton(type):
     _instances = {}
     def __call__(cls, *args, **kwargs):
-        if cls not in cls._instances:
-            cls._instances[cls] = super(EnvironmentsSingleton, cls).__call__(*args, **kwargs)
-        return cls._instances[cls]
+        repo_path = kwargs['repo_path']
+        if repo_path not in cls._instances:
+            cls._instances[repo_path] = super(EnvironmentsSingleton, cls).__call__(*args, **kwargs)
+        return cls._instances[repo_path]
 
 
 class Environments(metaclass=EnvironmentsSingleton):
--- a/src/hangar/dataset.py
+++ b/src/hangar/dataset.py
@@ -79,7 +79,7 @@ class DatasetDataReader(object):
         self._index_expr_factory.maketuple = False
         self._schema_dtype_num = varDtypeNum
         self._mode = mode
-        self._fs = FileHandles()
+        self._fs = FileHandles(repo_path=repo_pth)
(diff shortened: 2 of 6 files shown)


src/hangar/context.py
```diff
@@ -175,9 +175,10 @@ def abort_reader_txn(self, lmdbenv: lmdb.Environment) -> bool:
 class EnvironmentsSingleton(type):
     _instances = {}
     def __call__(cls, *args, **kwargs):
-        if cls not in cls._instances:
-            cls._instances[cls] = super(EnvironmentsSingleton, cls).__call__(*args, **kwargs)
-        return cls._instances[cls]
+        repo_path = kwargs['repo_path']
```
lantiga: What if `repo_path` is not specified during construction?

rlizzo:
So this is an interesting point. Ive been purposefully avoiding putting a close method on the repo because i think it would be annoying for users to have to remember. Only checkouts need to be closed, which should not close the entire repo context when they are called. 

Not sure there's a solution if in stuck on not putting in an explicit `repo.close()` method?

src/hangar/context.py
rlizzo: This has been fixed with the validator defined here:

https://github.com/tensorwerk/hangar-py/pull/20/files#diff-0bed44456b6a0cd62fbeaca8f6ab20aeR136

rlizzo:
Closes #5 and #4

src/hangar/context.py
lantiga: However here we assume that `kwargs` contains `repo_path`: I know it's not user-facing, but we need to throw a properly informative exception here.

lantiga:
I think we should at least give the option of cleaning up the singletons (either a specific one or all of them). In general I'm not 100% sure why we need a singleton (or something like it) design-wise, can you explain in a one-liner?

PR #22, opened 2019-05-01
Core concepts documentation

rlizzo:
High level overview of hangar concepts, problems solved, and how it works. 

Ping for discussions, edits, and todo's @lantiga


rlizzo:
Need to squash commits before sending PR.

PR #23, opened 2019-05-01
Added documentation discussing hangar core concepts

rlizzo:
Discusses core data model, problems we try to solve, and the rationale
for solving them in the first place

ping @lantiga for discussion and review.

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -225,7 +225,7 @@ a single backend (each of which will have significant performance tradeoffs for
 arrays of particular datatypes, shapes, and access patterns), we simultaneously
 store different data pieces in the backend which is most suited to it. A great
 deal of care has been taken to optimize parameters in the backend interface
-which affecting performance and compression of data samples.
+which affects performance and compression of data samples.
 
 The choice of backend to store a piece of data is selected automatically from
 heuristics based on the dataset specification, system details, and context of
@@ -291,6 +291,14 @@ the same.
 How To Overcome The "Size" Problem
 ----------------------------------
 
+.. note::
+
+    The features described in this section are in active development for a
+    future release. This is an exciting feature which we hope to do much more
+    with; Time is our main constraint right now. Hangar is a young project, and
+    is rapidly evolving. Current progress can be tracked in the `GitHub
+    Repository <https://github.com/tensorwerk/hangar-py>`_
+
 Even if the greatest tool imaginable existed to version, branch, and merge
 datasets, it would face one massive problem which if it didn't solve would kill
 the project: *The size of data can very easily exceeds what can fit on (most)
@@ -321,12 +329,12 @@ client/server.
     the following example may provide some insight into the implications of this
     property:
 
-        If you ``clone`` some hangar repository, Bookeeping says that "some number
-        of data piece exist" and they should retrieved from the server. However,
-        the bookeeping records transfered in a ``fetch`` / ``push`` / ``clone``
-        operation do not include information about where that piece of data
-        existed on the client (or server) computer. Two synced repositories can
-        use completly different backends to store the data, in completly
+        If you ``clone`` some hangar repository, Bookeeping says that "some
+        number of data pieces exist" and they should retrieved from the server.
+        However, the bookeeping records transfered in a ``fetch`` / ``push`` /
+        ``clone`` operation do not include information about where that piece of
+        data existed on the client (or server) computer. Two synced repositories
+        can use completly different backends to store the data, in completly
         different locations, and it does not matter - Hangar only guarrentees
         that when collaborators ask for a data sample in some checkout, that
         they will be provided with identical arrays, not that they will come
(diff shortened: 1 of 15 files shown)


rlizzo:
ping @hhsecond for review as well.

docs/concepts.rst
```diff
@@ -0,0 +1,448 @@
...
+
+We will show how hangar solves these questions in a high-level guide below.
+For a deep dive into the Hangar internals, we invite you to check out the
+:ref:`ref-hangar-under-the-hood` page.
+
+****************************
+How Hangar Thinks About Data
+****************************
+
+From this point forward, **when we talk about "Data" we are actually talking
```
elistevens: I would make a heading of `Abstraction 0: What is "data"?` and be explicit about what numeric types are supported.

Questions I had when reading this include:

- Will my data be preserved bit-exact?
- What about complex numbers?
- Will it support some weird, custom fixed-point format I might want to use?

I think that listing the subset of numpy types supported will answer most of this.

docs/concepts.rst
```diff
@@ -0,0 +1,448 @@
...
+about n-dimensional arrays of numeric information. To Hangar, "Data" is just a
+collection of numbers being passed into and out of it.** Data does not have a
+file type, it does not have a file-extension, it does not mean anything to
+Hangar itself - it is just numbers. This theory of "Data" is nearly as simple as
+it gets, and this simplicity is what enables us to be unconstrained as we build
+abstractions and utilities to operate on it.
+
+Abstraction 1: What is a Repository?
```
elistevens: Move this later, since you use `Dataset` here, but we don't know what it is yet.  I'd start with `Sample` then `Dataset` then `Repository` since that builds from small to large.

docs/concepts.rst
```diff
@@ -0,0 +1,448 @@
...
+High Performance From Simplicity
+--------------------------------
+
+Because Hangar is designed to deal (almost exclusively) with numerical arrays,
+we are able to "stand on the shoulders of giants" once again by utilizing many
+of the well validated, highly optimized, and community validated numerical array
+data management utilities developed by the High Performance Computing community
+over the past few decades.
+
+In a sense, the backend of Hangar servers two functions:
```
elistevens: "serves"

docs/concepts.rst
```diff
@@ -0,0 +1,448 @@
...
+
+1) Bookkeeping: recording information about about datasets, samples, commits, etc.
+2) Data Storage: highly optimized interfaces which store and retrieve data from
+   from disk through it's backend utility.
```
elistevens: "its"

docs/concepts.rst
rlizzo: Added!

rlizzo:
Thanks @elistevens, you're comments have been addressed! 

ping @lantiga for general comments before I proofread and merge.

docs/concepts.rst
```diff
@@ -0,0 +1,470 @@
...
+The details are explained much more thoroughly in :ref:`ref-hangar-under-the-hood`.
+
+Because Hangar only considers data to be numbers, the choice of backend to store
+data is (in a sense) completely arbitrary so long as ``Data In == Data Out``.
+**This fact has massive implications for the system**; instead of being tied to
+a single backend (each of which will have significant performance tradeoffs for
+arrays of particular datatypes, shapes, and access patterns), we simultaneously
+store different data pieces in the backend which is most suited to it. A great
+deal of care has been taken to optimize parameters in the backend interface
+which affecting performance and compression of data samples.
```
elistevens: "which affecting" review wording?

docs/concepts.rst
```diff
@@ -0,0 +1,470 @@
...
+    Hangar and stored immutably once committed. However, the access patern, backend,
+    and locating information for this data piece may (and over time, will) be unique
+    in every hangar repository instance**.
+
+    Though the details of how this works is well beyond the scope of this document,
+    the following example may provide some insight into the implications of this
+    property:
+
+        If you ``clone`` some hangar repository, Bookeeping says that "some number
+        of data piece exist" and they should retrieved from the server. However,
```
elistevens: "pieces"?

docs/concepts.rst
```diff
@@ -0,0 +1,470 @@
...
+release will even include the ability to stream data directly to a hangar
+checkout and materialize the data in memory without having to save it to disk at
+all!
+
+More importantly: **Since Bookkeeping describes all history, merging can be
+performed between branches which may contain partial (or even no) actual data**.
+Aka. You don't need data on disk to merge changes into it. It's an odd concept
+which will be explained more in depth in the future.
+
+.. note::
```
elistevens: Move this up for full transparency.

docs/concepts.rst
```diff
@@ -0,0 +1,470 @@
...
+    [[0, 1, 2],        [[0, 1, 2],
+     [0, 1, 2], ----->  [2, 2, 2], ----- MERGE CONFLICT
+     [0, 1, 2]]         [3, 3, 3]]      /
+          \                            /
+           \            commit C      /
+            \                        /
+             \          [[1, 1, 1], /
+              ------->   [0, 1, 2],
+                         [0, 1, 2]]
+
```
elistevens: Needs a bit of wrap-up saying "so the merge commit author must pick one, or provide a new array" or w/e.

elistevens: Also maybe a quick word about what happens with metadata and merges (I assume the union of added keys gets added, the union of removed keys gets removed, and conflicts are generated on anything touching the same key in different ways or with different values?).

rlizzo:
Thanks for the detailed feedback @elistevens. All comments have been resolved and I am going to merge this now

Issue #24, opened 2019-05-02
Environment Singleton Needed?


rlizzo:
I think we should at least give the option of cleaning up the singletons (either a specific one or all of them). In general I'm not 100% sure why we need a singleton (or something like it) design-wise, can you explain in a one-liner?

_Originally posted by @lantiga in https://github.com/tensorwerk/hangar-py/pull/20#issuecomment-488382678_

rlizzo:
From #21:

> With the bugfixes to #5 #4 in PR: #20, environment handle singletons can now potentially live indefinitly as there is no cleanup method.
> 
> Is there a way to GC the objects without introducing an explicit `.close()` method on the repository?
> 
> Tag @lantiga for discussion here.

lantiga:
@rlizzo is your latest WIP making this one obsolete?

rlizzo:
Yup! This will be gone as well as the HDF5 Singleton. 

Linking this to #70 for future reference.

rlizzo:
Fixed in #70

Issue #21, opened 2019-05-01
Singletons live indefinitly


rlizzo:
With the bugfixes to #5 #4 in PR: #20, environment handle singletons can now potentially live indefinitly as there is no cleanup method. 

Is there a way to GC the objects without introducing an explicit `.close()` method on the repository?

Tag @lantiga for discussion here.

rlizzo:
This is a duplicate of #24.

PR #25, opened 2019-05-02
Checkout testcases

hhsecond:
## Description
Test cases & name validity check for the branch name

## Types of changes

## Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -7,6 +7,7 @@ import numpy as np
 from hangar import Repository
 import hangar
 
+
 @pytest.fixture(autouse=True)
 def reset_singletons(monkeypatch):
     '''
--- a/tests/test_checkout.py
+++ b/tests/test_checkout.py
@@ -2,6 +2,7 @@ import numpy as np
 import pytest
 import shutil
 import platform
+from hangar.records import summarize
 
 
 class TestCheckout(object):
@@ -9,7 +10,7 @@ class TestCheckout(object):
     def test_two_write_checkouts(self, repo):
         w1_checkout = repo.checkout(write=True)
         with pytest.raises(PermissionError):
-            w2_checkout = repo.checkout(write=True)
+            repo.checkout(write=True)
         w1_checkout.close()
 
     def test_two_read_checkouts(self, repo, array5by7):
@@ -17,21 +18,29 @@ class TestCheckout(object):
         dataset_name = 'dset'
         r_ds = w_checkout.datasets.init_dataset(name=dataset_name, prototype=array5by7)
         r_ds['1'] = array5by7
+        w_checkout.metadata.add('init', 'array5by7 added')
         w_checkout.commit('init')
         r1_checkout = repo.checkout()
         r2_checkout = repo.checkout()
         assert np.allclose(r1_checkout.datasets['dset']['1'], array5by7)
-        assert np.allclose(r1_checkout.datasets['dset']['1'], r2_checkout.datasets['dset']['1'])
+        assert np.allclose(
+            r1_checkout.datasets['dset']['1'], r2_checkout.datasets['dset']['1'])
+        assert r1_checkout.metadata.get('init') == 'array5by7 added'
+        assert r2_checkout.metadata.get('init') == 'array5by7 added'
         r1_checkout.close()
         r2_checkout.close()
         w_checkout.close()
 
-    def test_write_with_r_checkout(self, written_repo, array5by7):
+    @pytest.mark.skip(reason='Need to Fix TODO')
+    def test_write_with_read_checkout(self, written_repo, array5by7):
         co = written_repo.checkout()
+        # TODO: This should raise more informative error message
+        # Now it's type error because `init_dataset` is None
         with pytest.raises(TypeError):
             co.datasets.init_dataset(name='dset', shape=(5, 7), dtype=np.float64)
-            co.datasets['_dset']['1'] = array5by7
-            co.datasets['_dset'].add('1', array5by7)
+        # TODO: This should be permission error
+        with pytest.raises(AttributeError):
+            co.metadata.add('a', 'b')
         co.close()
 
     @pytest.mark.skipif(platform.system() == 'Windows',
@@ -41,9 +50,9 @@ class TestCheckout(object):
         shutil.rmtree(written_repo._repo_path)
         with pytest.raises(OSError):
             co.datasets.init_dataset('dset', prototype=array5by7)
+        co.metadata.add('a', 'b')
         dset = co.datasets['_dset']
         dset['1'] = array5by7
-        # todo: shouldn't it be atleast same as above
         with pytest.raises(FileNotFoundError):
             co.commit()
         co.close()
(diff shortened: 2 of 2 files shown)


src/hangar/repository.py
```diff
@@ -651,6 +651,12 @@ def create_branch(self, branch_name, base_commit=None):
         bool
             if the operation was successful.
         '''
+        if not is_ascii_alnum(branch_name):
```
hhsecond: I could move this to a separate PR if you want to keep this PR just for test cases

tests/conftest.py
```diff
@@ -6,7 +6,11 @@
 
 from hangar import Repository
 import hangar
+from hangar.records import summarize
 
+
+# TODO: IMO, we should not have this cleanup and let tests catch issues
+# if there are any
```
rlizzo: Can you expand on this? 

I believe it is actually required, since without explicitly cleaning up the past instances, when a new repository is initiated with the same python process, the old state would still be left in the singleton instance dicts. This was the fix which made tests actually work in the first place. Unless https://github.com/tensorwerk/hangar-py/pull/20 fixed this behavior (I'm not sure, haven't tested it myself) then I think it's still needed? Please correct me if I'm wrong though!

tests/test_checkout.py
```diff
@@ -8,64 +8,207 @@ class TestCheckout(object):
...
-
-    def test_merge_multiple_checkouts(self):
-        pass
-
-    def test_merge_diverged(self):
-        pass
-
-    def test_merge_diverged_conflict(self):
-        pass
+    def test_operate_on_dataset_after_closing_old_co(self, repo, array5by7):
```
rlizzo: style note: don't use abreviations in test case function names. I don't care if test case function names go over some arbitrary line length limit (90ish for the main code base) since they should be as descriptive as possible.

tests/test_checkout.py
```diff
@@ -8,64 +8,207 @@ class TestCheckout(object):
...
-        assert np.allclose(r1_checkout.datasets['dset']['1'], r2_checkout.datasets['dset']['1'])
+        assert np.allclose(
+            r1_checkout.datasets['dset']['1'], r2_checkout.datasets['dset']['1'])
+        assert r1_checkout.metadata.get('init') == 'array5by7 added'
+        assert r2_checkout.metadata.get('init') == 'array5by7 added'
         r1_checkout.close()
         r2_checkout.close()
         w_checkout.close()
 
     def test_write_with_r_checkout(self, written_repo, array5by7):
```
rlizzo: style note: no abbreviations in test case function names. 

NOTE: this applies for all test functions, no need for me to clutter the review with comments for every single one.

tests/test_checkout.py
```diff
@@ -8,64 +8,207 @@ class TestCheckout(object):
...
-
-    def test_merge_diverged_conflict(self):
-        pass
+    def test_operate_on_dataset_after_closing_old_co(self, repo, array5by7):
+        co = repo.checkout(write=True)
+        dset = co.datasets.init_dataset('dset', prototype=array5by7)
+        co.commit()
+        co.close()
+        co = repo.checkout(write=True)
+        # TODO: what is the design here. IMO, the old instance should not execute
```
rlizzo: **It absolutely SHOULD NOT WORK! Does it?? That's a huge bug if so!!**

tests/test_checkout.py
```diff
@@ -8,64 +8,207 @@ class TestCheckout(object):
...
+        dset.add(array5by7, '1')
+        co.commit()
+        co.close()
+        # TODO: even after the close, I could still fetch the
```
rlizzo: This should not work. #

tests/test_checkout.py
```diff
@@ -8,64 +8,207 @@ class TestCheckout(object):
...
+
+    def test_invalid_name(self, repo):
+        with pytest.raises(ValueError):
+            repo.create_branch('dummy branch')
+
+    def test_new_branch_from_new_repo(self, repo):
+        branch = repo.create_branch('testbranch')
+        # TODO: it doesnt raise value error to the user
+        # instead return false and print out the value error
+        # Ideally, it should let users create new branch
```
rlizzo: Nope. For now, I want users to explicitly say they want to create a branch. 

To better explain the reasoning here: I'm hard set on being explicit for branch operations because the model is slightly different from `Git`'s in the sense that there is no way to do a `write-enabled checkout` on a `detached HEAD commit`. All writes have to be performed on the `"tip" HEAD commit` of a branch. Without being explicit about the branch creation and checkout operations, this subtle difference might lead to confusion in day-to-day use.

tests/test_checkout.py
```diff
@@ -8,64 +8,207 @@ class TestCheckout(object):
...
+        assert type(branch) is str
+        co = written_repo.checkout(write=True, branch_name=branch)
+        assert co._branch_name == branch
+        co.datasets['_dset']['1'] = array5by7
+        co.metadata.add('a', 'b')
+        co.commit()
+        co.close()
+        written_repo.merge('test merge', 'master', branch)
+        co = written_repo.checkout()
+        # TODO: It's good to give access to name of the branch from Readcheckout
```
rlizzo: I agree in principle, but disagree in practice. Since a read checkout and write checkout can operate at the same time on the same repository, we should only provide the commit which the reader checkout is operating on, not the actual branch name. 

Imaging the following scenario where a reader checks out `branch "A"` which (at the time of checkout) has head commit `2`. 

```
                         Reader c/o 
0 ---------> 1
              \___________> 2
                         branch A
```

Then a writer check's out `branch "A"` as well

```
                           Reader c/o
0 ---------> 1             Writer c/o
              \___________> 2
                         branch A
```

And then (while the reader is still active) the writer makes `commit 3` on `branch "A"`

```
                        Reader c/o
0 ---------> 1                                next writer c/o
            \___________-> 2 _________________-> 3
                        branch A
```

If we provided the "branch name" to the reader checkout object, it would be defined when it was first created, but then could potentially change to a `null` value over the course of it's lifetime... this would be a crazy behavior for user defined code which might depend on a branch name.

If we only provide the commit hash, we can be sure that it will never change under any circumstances, and a user can actually write application code which depends on that value for whatever need they might have.

tests/test_checkout.py
```diff
@@ -8,64 +8,207 @@ class TestCheckout(object):
...
+        assert (co.datasets['_dset']['1'] == array5by7).all()
+        assert co.metadata.get('a') == 'b'
+
+    def test_merge_without_closing_previous_checkout(self, written_repo, array5by7):
+        branch = written_repo.create_branch('testbranch')
+        co = written_repo.checkout(write=True, branch_name=branch)
+        co.datasets['_dset']['1'] = array5by7
+        co.commit()
+        # TODO: Need better error message
+        # lockHeld check at line #235 of diff.py never happens
```
rlizzo: good catch. please file a bug and link it here.

tests/test_checkout.py
```diff
@@ -8,64 +8,207 @@ class TestCheckout(object):
...
+            written_repo.merge('test merge', 'master', branch)
+
+    def test_name_conflict(self, written_repo, array5by7):
+        written_repo.create_branch('testbranch')
+        with pytest.raises(ValueError):
+            written_repo.create_branch('testbranch')
+
+    def test_delete_branch(self, written_repo):
+        with pytest.raises(NotImplementedError):
+            # TODO should raise "branch doesn't exist error"
```
rlizzo: I haven't figured out how we should handle removing branches yet. the logic it a bit complicated because we need to figure out what data is referenced in other commits, and which can be safely removed from disk. please create an issue and link it here for this.

tests/test_checkout.py
```diff
@@ -8,64 +8,207 @@ class TestCheckout(object):
...
+        written_repo.merge('commit message', 'master', branch2)
+        co = written_repo.checkout()
+        print(co.datasets['_dset']['1'])
+        print(co.metadata.get('a'))
+
+    def test_new_branch_from_where(self, written_repo, get_history, array5by7):
+        branch1 = written_repo.create_branch('testbranch1')
+        branch2 = written_repo.create_branch('testbranch2')
+        co1 = written_repo.checkout(write=True, branch_name=branch1)
+        # TODO: better way to get history: could be part of checkout itself
```
rlizzo: agreed. please create issue and link it here.

tests/test_checkout.py
```diff
@@ -8,64 +8,207 @@ class TestCheckout(object):
...
+        co.commit()
+        co.close()
+        co = written_repo.checkout(write=True, branch_name=branch2)
+        newarray = np.zeros_like(array5by7)
+        co.datasets['_dset']['1'] = newarray
+        co.metadata.add('a', 'c')
+        co.commit()
+        co.close()
+        written_repo.merge('commit message', 'master', branch1)
+        # TODO: This should raise a conflict coz of same key
```
rlizzo: I'm not following, can you explain what the bug is a bit better?

tests/conftest.py
```diff
@@ -69,3 +73,12 @@ def randomsizedarray():
     a = random.randint(2, 8)
     b = random.randint(2, 8)
     return np.random.random((a, b))
+
+
+@pytest.fixture()
+def get_history(scope='sesssion'):
```
rlizzo: Please don't define this as a fixture function, it's a built in utility of the library, and should just be called directly from a repository object in any tests which need it via the method `repo.summary()` (optionally with `return_contents=True` for programatic evaluation.  No need to make test fixtures more complex then they already area :)

tests/test_checkout.py
hhsecond: in both branch1 and branch2, test case has `_dset` that has data on the key `1`. It is essentially a merge conflict while merging but what hangar currently does is overwrite.

tests/conftest.py
hhsecond: I was thinking we'll keep all the dependencies to fixtures and keep the `import`s minimal as possible (Just another way of test designs we used to follow). Let's do it your way

tests/test_checkout.py
rlizzo: damn ok. big bug here!

tests/conftest.py
hhsecond: Exactly, By cleaning up the past instances, we are avoiding a chance to find whether we have that issue fixed or not. I understand it is required if it is not fixed (for other tests to run) but my point is, we need to remember to remove this fixture once we confirm that particular issue is gone.

tests/test_checkout.py
hhsecond: created #36

tests/test_checkout.py
hhsecond: created #36

tests/test_checkout.py
hhsecond: created #37

tests/test_checkout.py
hhsecond: #38

tests/test_checkout.py
hhsecond: #39

tests/test_checkout.py
hhsecond: `repo.log(return_contents=True)` does this, right?

tests/test_checkout.py
hhsecond: This makes a lot of sense. Although, I have a slightly convoluted solution where we raise a warning if the head moved further when the user tries to fetch the branch name from a read checkout and return the branch name. If the head not moved, return the branch name without raising the warning. I am just thinking aloud and perhaps this could be a problem for another day.

tests/test_checkout.py
hhsecond: Correct me If I am wrong; What I understand from the context above is, on a new repository user should not be able to create a new branch until the first commit?

tests/conftest.py
rlizzo: Ok. Linking this discussion to #24 and #21 since they will decide how this is handled in the long term.

tests/test_checkout.py
rlizzo: So I double checked, and this information is already provided to each of the checkout objects in one of their underscore attributes:

https://github.com/tensorwerk/hangar-py/blob/065ae0735c630f627fb5f5d9cd634739408a5733/src/hangar/checkout.py#L42

As part of #32 I'll make a `commit` attribute available to the user (following the descriptor protocol to prevent a user setting this value to something else after construction)

I'll do the same for the `branch_name` attribute for a write-enabled checkout.

https://github.com/tensorwerk/hangar-py/blob/065ae0735c630f627fb5f5d9cd634739408a5733/src/hangar/checkout.py#L124

tests/test_checkout.py
rlizzo: yeah!

tests/test_checkout.py
hhsecond: @rlizzo I am good to merge the PR once this conversation can be resolved (either through a Fix or an issue that tracks this test case). Let me know your thoughts. As a first-time hangar user, I would expect hangar to let me create a branch on an empty repository

rlizzo:
So I'm a bit confused as to the status of this? I think there's a lot of overlap with #32, and this branch doesn't seem like it's really kept up with the changes in master. 

Do you want to continue with this PR? Or include the relevant changes in #32 along with a most focused commit for the remainder?

hhsecond:
The TODOs in this PR is not just related to #32. There are few other changes that we haven't looked into at all. So I would say, let's merge this and then along with rebasing #32, will sort out what's required there. Once #32 lands, we'll have a stable master on which I can do a clean up perhaps.

rlizzo:
Ok sounds good. I'm just going to rerun the tests to make sure everything is OK, then will merge this and rebase #32.

PR #31, opened 2019-05-06
Multi-add docstring addition

rlizzo:
added docstring to the multi-add mapping funciton. changed argument name to be less verbose

Fixes #30

PR #34, opened 2019-05-07
Quick Fix for variable name propagation issue in #31

hhsecond:
## Description
Change of variable name in [here](https://github.com/tensorwerk/hangar-py/blob/044937c5a00fa48bd2889cce878648404645591d/src/hangar/dataset.py#L743) did not go throughout the function and we are using the old name

## Types of changes

## Checklist:

### DIFF
--- a/src/hangar/dataset.py
+++ b/src/hangar/dataset.py
@@ -783,13 +783,13 @@ class Datasets(object):
                 self._datasets[dskey] = self._datasets[dskey].__enter__()
 
         try:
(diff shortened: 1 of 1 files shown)


hhsecond:
This type of simple issue is extremely unnoticeable & easily sneak in through the person who develops but almost always gets caught by the reviewer. Let's implement no-merge-without-one-approval policy, perhaps?

rlizzo:
Good catch! 

I think we're probably at the point where we should do mandatory reviews. Why don't we set the 0.1.0 release as the official start of that policy?

Issue #36, opened 2019-05-07
Dataset object is operatable even after closing the checkout


hhsecond:
## Expected Behavior
After closing the checkout, the dataset object should not be operatable. We need to decide what all are possible on the dataset object though. So far, we know
1. User should not be able to add data.
2. User should not be able to fetch data

## Current Behavior
1.
```python
co = repo.checkout(write=True)
dset = co.datasets.init_dataset('dset', prototype=array5by7)
co.commit()
co.close()
co = repo.checkout(write=True)  # without opening the checkout again, user won't be able to do the addition even now
dset.add(array5by7, '1')  # user can add data to dset object
```
2.
```python
co = repo.checkout(write=True)
dset = co.datasets.init_dataset('dset', prototype=array5by7)
dset.add(array5by7, '1')
co.commit()
co.close()
print(dset['1'])  # user can fetch data from dset object
```

These issues are part of #25 
- https://github.com/tensorwerk/hangar-py/pull/25#discussion_r281058557
- https://github.com/tensorwerk/hangar-py/pull/25#discussion_r281058815

rlizzo:
So the problem with this is that it's stupidly simple to fix, but adds more the the increasing list of verifications which need to occur in order to get or add data to a dataset/metadata object... we need some benchmarks soon before we "verify" ourselves into a tight spot here. 

Note: Most of the explanation below is for `write-enabled` checkouts, since that is the really dangerous case where this could go wrong, a `read only` checkout could only ever just act odd, not actually cause harm to the underlying records. 

Since a user is provided direct access to the dataset objects, and are allowed to keep references to them outside the hangar namespace (and checkout object from which they are constructed), closing a checkout currently only `"self destructs"` the ability for new `dataset`/`metadata` object access. See:

https://github.com/tensorwerk/hangar-py/blob/065ae0735c630f627fb5f5d9cd634739408a5733/src/hangar/checkout.py#L158-L177

Since the repo object has no way to communicate with reviously created `dataset` or `metadata` object to tell them they need to stop working, the fix would be to implement checks in the object themselves. By passing the `__writer_lock` key and `_branchenv` environment down the stack, the objects could in theory ask for permission to operate every time they want to do something (like the checkout object does). 

[...]

PR #41, opened 2019-05-07
Invalidate existing dataset or metadata accessor objects after a checkout is closed

rlizzo:
These changes ensure that when a checkout (reader or writer) is closed, that any references to `dataset` or `metadata` objects are invalidated and will not perform any type of operation. 

## Description

Instead of explicitly validating that some sort of `IS_ALIVE` sentinal on every call to the `dataset` or `metadata` interactor objects, we  use an invalidation scheme based on weakrefs. The control flow is as follows: 

1) A strong reference to a dataset or metadata instance is created and held in a "`_private`" attribute of the `checkout` class. In the new scheme, this is the ONLY strong reference which will be kept to the members. 

2) For public consumption, `datasets` and `metadata` methods are defined in the `checkout` object using the `property` protocol to remap the `getter` to the internal attributes. Because no `setter` is defined, these properties cannot be changed by the user.

3) At this point, we do a "bait-and-switch", providing a `weakref proxy` to the requested object. By definition, these do not increment the reference count, but pass through all calls to the checkout's instance of the actual object.

4) when the `checkout.close()` method is called, the checkout objects `._datasets` and `._metadata` attributes are deleted and we explicitly perform a garbage collection.

5) Since `weakrefs` don't increment reference counts, when the attributes are deleted on close and gc is run, the `proxy` mapping no longer has a valid reference. It will raise a `ReferenceError()` if any operation is attempted on it. 

6) Validations performed in the `checkout` class ensure that the object is essentially dead to the user, and a new checkout must be created to interact with the repository again. 

This is a performant and clean way of ensuring that users are not able to perform operations during a time they shouldn't be able to. Methods are implemented for both read and write checkouts and both function as expected and pass all tests. 

## Motivation and Context

Examples of the problem before this fix were first shown in Issue #36. Please refer to that posting for further background. 

## How Has This Been Tested?

Test cases have been added to the test suite.

## Types of changes

## Checklist:

Please review @hhsecond and @lantiga

### DIFF
--- a/src/hangar/checkout.py
+++ b/src/hangar/checkout.py
@@ -46,19 +46,25 @@ class ReaderCheckout(object):
         self._labelenv = labelenv
         self._dataenv = dataenv
         self._hashenv = hashenv
-        self.metadata = MetadataReader(dataenv=self._dataenv, labelenv=self._labelenv)
-        self.datasets = Datasets._from_commit(
+        self._metadata = MetadataReader(dataenv=self._dataenv, labelenv=self._labelenv)
+        self._datasets = Datasets._from_commit(
             repo_pth=self._repo_path,
             hashenv=self._hashenv,
             cmtrefenv=self._dataenv)
 
     def _repr_pretty_(self, p, cycle):
+        '''pretty repr for printing in jupyter notebooks
+        '''
+        self.__verify_checkout_alive()
         res = f'\n Hangar {self.__class__.__name__}\
                 \n     Writer       : False\
-                \n     Commit Hash  : {self._commit_hash}\n'
+                \n     Commit Hash  : {self._commit_hash}\
+                \n     Num Datasets : {len(self._datasets)}\
+                \n     Num Metadata : {len(self._metadata)}\n'
         p.text(res)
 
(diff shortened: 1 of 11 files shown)


src/hangar/checkout.py
```diff
@@ -73,8 +139,12 @@ def close(self):
         system resources, which may improve performance for repositories with
         multiple simultaneous read checkouts.
         '''
-        for dsetHandle in self.datasets.values():
+        self.__verify_checkout_alive()
+        for dsetHandle in self._datasets.values():
             dsetHandle._close()
+        del self._datasets
+        del self._metadata
+        gc.collect()
```
lantiga: Before this goes in, let's try removing collect after `gc.disable()`, and see if the call to `del` is sufficient to invalidate the weak ref. I'd prefer if we got rid of the call to `collect`

rlizzo: So after a bit of testing it turned out this wasn't necessary after all! Nice spot!

rlizzo:
All review comments have been addressed. merging this now.

PR #42, opened 2019-05-07
Removed manual disk space freeing 

rlizzo:
which increased build times due to increase in disk quota from appveyor support

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -34,6 +34,7 @@ environment:
 init:
   - ps: echo $env:TOXENV
   - ps: ls C:\Python*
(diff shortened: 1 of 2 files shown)


rlizzo:
DO NOT MERGE - AWAITING RESOLUTION FROM APPVERYOR SUPPORT

rlizzo:
APPVEYOR SUPPORT was able to increase disk usage. Paired with turning down the amount of memory we request for each lmdb environment (from 10GB to 5GB) this works much faster. The reduced db size shouldn't be an issue at all...

PR #50, opened 2019-05-15
Quick fix for substring check

### LINKED ISSUE #49: [BUG REPORT] Traversing through dataset data records is buggy
**Describe the bug**
The function `RecordQuery._traverse_dataset_data_records(self, dataset_name)` is has a critical bug which related to the condition check we are doing to make sure we are checking for the provided dataset (This must be same with `_traverse_dataset_schema_records` function as well since the logic is similar there as well). If the repository has multiple datasets where one's name has another dataset's name as a substring in the start of the name, would cause the condition check `dataRecKey.startswith(startDatasetRecordCountRangeKey)` to return wrong data.

**Severity**

Select an option:
- [ ] Data Corruption / Loss of Any Kind
- [x] Unexpected Behavior, Exceptions or Error Thrown
- [ ] Performance Bottleneck

**To Reproduce**
First code snippet to save two datasets in the same repository named `MNIST` and `MNISTLabel`

```python
import os
from hangar import Repository
import torchvision
import numpy as np

data = torchvision.datasets.MNIST('.', download=True)
path = os.path.dirname(os.path.abspath(__file__))
repo = Repository(path=path)
repo.init(user_name='hhsecond', user_email='sherin@tensorwerk.com', remove_old=True)
co.datasets.init_dataset('MNIST', shape=(28, 28), dtype=np.uint8)
co.datasets.init_dataset('MNISTLabel', shape=(1,), dtype=np.int64)
mnist_dset = co.datasets['MNIST']
label_dset = co.datasets['MNISTLabel']

with mnist_dset, label_dset:
    for i, (image, label) in enumerate(data):
        print(i)
        mnist_dset[str(i)] = np.array(image)
        label_dset[str(i)] = np.array([label])
co.commit('0.1')
co.close()
```

The second part where we could try to check the name of dataset `MNIST`
```python
co = repo.checkout()
mnist = co.datasets['MNIST']
mnistlabel = co.datasets['MNISTLabel']
print(len(mnist))
```
`len` on dataset triggers the above functions which would cause it to fail

**Expected behavior**
- In the above case, the user should get the actual length
- Generally, `_traverse_dataset_data_records` should return a dictionary of data records with name as key and hash as value

hhsecond:
## Motivation and Context
Quick fix for substring check to fix #49 

## Description
Substring check for dataset name dataset traversal now checks for a substring that ends with `:` which prevents from accidentally including other datasets that have the same name in the start of the datasets name
eg: 
`MNIST` and `MNISTLabel`

## Types of changes

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:

### DIFF
--- a/src/hangar/records/queries.py
+++ b/src/hangar/records/queries.py
@@ -2,6 +2,8 @@ from . import parsing
 from .. import config
 from ..context import TxnRegister
 
+SEP = config.get('hangar.seps.key')
+
 
 '''
 Data record queries
@@ -110,10 +112,11 @@ class RecordQuery(object):
             with datatxn.cursor() as cursor:
                 # with datatxn.cursor() as cursor:
                 dataRecordsExist = cursor.set_range(startDatasetRecordCountRangeKey)
+                dataRecKeySubString = f'{startDatasetRecordCountRangeKey.decode()}{SEP}'.encode()
                 cursor.next()
                 while dataRecordsExist:
                     dataRecKey, dataRecVal = cursor.item()
-                    if dataRecKey.startswith(startDatasetRecordCountRangeKey):
+                    if dataRecKey.startswith(dataRecKeySubString):
                         data_records[dataRecKey] = dataRecVal
                         dataRecordsExist = cursor.next()
                         continue


hhsecond:
@rlizzo 
This is a quick fix for the issue #49. If this solution is good enough, I'll go ahead and add test cases for the same before merging the PR. Just wanted to make sure there are no logical changes on the way that rewrites those functions entirely.

src/hangar/records/queries.py
```diff
@@ -110,10 +110,11 @@ def _traverse_dataset_data_records(self, dataset_name):
             with datatxn.cursor() as cursor:
                 # with datatxn.cursor() as cursor:
                 dataRecordsExist = cursor.set_range(startDatasetRecordCountRangeKey)
+                dataRecKeySubString = f'{startDatasetRecordCountRangeKey.decode()}:'.encode()
```
rlizzo: Instead of explicitly specifying the "colon" `:` as the seperator, we have defined the "seperator" constant in the config file here:

https://github.com/tensorwerk/hangar-py/blob/dc08edb25c66ea9b2d029ac8ba8c771de2b18bbb/src/hangar/config_hangar.yml#L14-L15

At the top of the file, define:

```python
SEP = config.get('hangar.seps.key')
```

like we do in the `parsing.py` file here:

https://github.com/tensorwerk/hangar-py/blob/dc08edb25c66ea9b2d029ac8ba8c771de2b18bbb/src/hangar/records/parsing.py#L21-L26

Which would change the addition to:

```python
 dataRecKeySubString = f'{startDatasetRecordCountRangeKey.decode()}{SEP}'.encode()
```

The better solution (but one which may require some debuging work) would probably be to add the `SEP` to the end of the `dset data count` key which is incred/decred every time a sample is added/removed. 

This would be in the code here (and the reverse code as well)

https://github.com/tensorwerk/hangar-py/blob/dc08edb25c66ea9b2d029ac8ba8c771de2b18bbb/src/hangar/records/parsing.py#L286-L289

Remember to look at the `repo._details()` call output to see this for yourself. 

image

However, the base solution does work with that modification!

hhsecond: Let me make the tweak and add test cases as soon as I could make some time. Speaking of dset count key, a completely unrelated issue, is there any reason we are looping over the whole dataset samples when we do `len(dataset)`. We could just fetch this key and return the count?

rlizzo: I had previously had some issues with the count being inacurate in certain scenarious and that method of counting was a quick hack so i could get the rest of the system working... however those issues have long been fixed, and this is just a remnant of that old hack.

 If you want to streamline that, it would be a welcome improvement!

hhsecond:
@rlizzo I changed the code so that `SEP` replaces `:`. One reason I found not to edit the `dataset_record_count_db_key_from_raw_key` function is, all of our `*_db_key_from_raw_key` has a similar structure -> returns the key without any separator. Changing that might bite us back later. What do you say? Anyway, this should fix our current issue.

rlizzo:
awesome. this works for me! merging now!

PR #67, opened 2019-05-27
Explicit verification of `write` argument

hhsecond:
## Motivation and Context
From a UX perspective, a user is tend to pass `branch_name` as positional argument to `checkout()`. Currently, `repo.checkout('branchname')` doesn't throw any error but takes the value `branchname` as value of `write` and falls to the `else` condition check inside the function. This creates subtle situations. An appropriate solution would be to make `branch_name` / `commit_hash` as the first argument but this might not be the usage that happens often. So for now, we can go with the check this PR introduces which is the check for the value of `write` explicitly.

## Description
Checking the value of `write` explicitly on `repo.checkout`

## Types of changes

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:

### DIFF
--- a/src/hangar/repository.py
+++ b/src/hangar/repository.py
@@ -142,6 +142,11 @@ class Repository(object):
             name parameter if it is set. Note: this only will be used in
             non-writeable checkouts, defaults to ''
 
+        Raises
+        ------
+        ValueError
+            If the value of `write` argument is not boolean
+
         Returns
         -------
         object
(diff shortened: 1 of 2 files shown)


hhsecond:
@rlizzo Done

rlizzo:
Great. Merging now!

Issue #13, opened 2019-04-26
Dataloaders for PyTorch


hhsecond:
@rlizzo I was thinking an API like `hangar.dataloaders.pytorch` (let me know if you have another structuring in your mind). Basically, the idea is to load data in batches synchronously or asynchronously and enable the features of PyTorch [DataLoader](https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader). My plan is to have a Dataset class and a DataLoader class which is essentially the way PyTorch's data loader work.

rlizzo:
So before we go ahead with this we should have a batched get method for the regular dataset method. If we get that right we can keep most of the functionality in the hangar base while making it really easy to plugin to the pytorch and tensorflow data loader framework

hhsecond:
@rlizzo I made the skeleton of the dataloader base class #73 . Adding features such as batching, shuffling to the base dataloader is not required as of now since both PyTorch and Tensorflow does that quite well. They just need an object from where they could get one sample at a time. In my view, the base class can be inherited and make advanced dataloaders inside hangar itself when it's required in the future.

rlizzo:
Hey @hhsecond it's going to be time to revisit this soon. I have a Dev branch where Read checkouts are thread and process safe (which includes a batched loader as well). I'm going to propose it in a PR in the next day or so, but we should coordinate on how this will plug in.

hhsecond:
Sure thing. I'll have a look into that PR before we could get into a call or something

hhsecond:
@rlizzo Do you have a preferred directory structure for including `dataloaders`? Also, I am thinking of the API as `hangar.dataloaders.TorchLoader`

rlizzo:
I don't have a preferred structure right now. Let's take a look a simple implementation and we can restructure as needed from there. 

I also like the `hangar.dataloaders.foo` for now. Since the use facing functions are very limited, we can probably just hoist the `TorchLoader` / `TensorflowLoader` functions up to the top level (`from hangar import TorchLoarder`) but that can happen with `__init__.py` magic later.

hhsecond:
If you want use init file for importing and stuff, we'll have to keep the data loader folder inside hangar folder (importing a level above the package is not recommended in python although we can work around it)

rlizzo:
Good point. We could get fancy with the packages spec in `setup.py`, but it's probably not worth the hastle.

[...]

Issue #71, opened 2019-05-30
[FEATURE REQUEST] Simple Command Line Interface ~~ hang ~~


rgalbo:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always frustrated when 

I want to use hanger (hang) at the terminal!

**Describe the solution you'd like**

I would like 

```
~> hang commit -m "update the training set"
```

```
~> hang merge < branches-of-interest>
```

**Describe alternatives you've considered**

**Additional context**

there are many simple cli for python. even the base cli tool making should be sufficient to not require any dependencies. I have no issue completing this myself in a new branch. Most likely could do so this afternoon with mild coaching. is there a slack or discord i should be aware of?

rlizzo:
Hi Rick, 

Thanks for the comment! 

Hangar actually already includes  a working (though very limited) CLI as of the latest (0.1.1) release. This should be working for both source builds & installations via `pip` / `conda`. You should be able to just run:

```
   $ hangar --help
```
from the command line to see the current abilities. 

As you've noticed, we haven't included any user-facing documentation of this feature. This is because it's not yet feature complete, nor is it under test-coverage. That said I use it fairly frequently without issue, and the design of Hangar makes the potential for bugs in this area fairly limited. What's there should be fully functional, but it's just not ready for general use. 

It would actually be really simple to flesh out the rest of the pertinent commands (`commit`, `merge`, `force-release-writer-lock`), which would actually make it useful. You can check out the CLI code in it's entirety in this file (it's really short):

https://github.com/tensorwerk/hangar-py/blob/master/src/hangar/cli.py

As you'll see, the commands are just `click` decorated functions which make calls straight to the same top level `Repository` class methods which are used in the programatic API. The `Repository` class is the entrance point for every user interaction with Hangar, no change required if it's on the command line . If you're curious, the `Repository` class defined in the following file:

https://github.com/tensorwerk/hangar-py/blob/master/src/hangar/repository.py

(Please excuse how ugly the fetch/push/clone methods are right now, the entire remote client / server implementation is about to undergo a major refactor in the next week or two; it'll be ready for the prime time in `v0.2` when we squash a few outstanding bugs and introduce partial clones). 

If you're interested in helping us out, we'd love the contribution in this area. I'm happy to help in any way I can!

[...]

PR #73, opened 2019-06-01
Dataloaders

hhsecond:
## Motivation and Context
Integration with other popular tools which could be used alongside hangar is really useful for users. Dataloaders is probably one of them.
#13 

## Description
Right now, it's just a skeleton to start the discussion about the structure. I'll change it as it moves forward.

## Types of changes

- [ ] Ready for review
- [x] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:

### DIFF
--- a/src/hangar/datautils/__init__.py
+++ b/src/hangar/datautils/__init__.py
@@ -1,0 +1,1 @@
+from .dataloader import DataLoader, TorchDataLoader, TensorflowDataLoader
--- a/src/hangar/datautils/dataloader.py
+++ b/src/hangar/datautils/dataloader.py
@@ -1,0 +1,25 @@
+class DataLoader(object):
+    '''
+    Base class for dataloaders. All the necessary APIs required for
+    a dataloader class must be created here. But the high level APIs
+    such as batching will be available in child classes. This class
+    must be inherited to create custom loaders for different integrations
+    as well such as PyTorch loader.
+    '''
+
+    def __init__(self, dataset):
+        pass
+
+    def __len__(self):
+        pass
+
+    def __getitem__(self, key):
+        pass
+
+
+class TorchDataLoader(DataLoader):
+    pass
+
+
+class TensorflowDataLoader(DataLoader):
+    pass


hhsecond:
@rlizzo I made the basic skeleton of the dataloader base class. For now, we could keep the other dataloaders also in the same file. Taking your discussion from #13, adding features such as batching, shuffling to the base dataloader is not required as of now since both PyTorch and Tensorflow does that quite well. They just need an object from where they could get one sample at a time. In my view, the base class can be inherited and make advanced dataloaders inside hangar itself when it's required in the future.

rlizzo:
Hey sherin I think we should put this on the back burner right now. The data loaders are important, but the bulk of the work to define multithreaded (multiprocess?) and batch reads actually should be implemented in the hangar core and enabled by default for some operations. These utilities could just implement the framework specific interfaces and make calls straight to the public hangar API. 

I'm also not thrilled about including them in the `hangar-py/src` directory. I would like to keep only User facing APIs and the backend processing workflows in there. Since these sorts of utility functions are just bolted on for convienence (similar to what we want to do with the CLI import/export) I'm thinking they should almost be stored as a separate module? 

What I mean is that we may want to split the `src` dir into `core` (containing what is in the repo right now) and a `convenience` dir (or some other similar name) to hold these. To keep a unified experience for importing the Hangar Repo or Dataloaders classes in a dependent codebases we could just raise the "apparent" level of each module to a unified interface. When used as intended (and not like most home brewed projects do) we can do some really useful things with a modules `__init__.py` file. I can handle all of that when it comes time...

With the changes introduced in #71, multithreaded reading got a whole lot simpler. Let's coordinate on those changes and how the backend now thinks about files before I leave for holiday on Tuesday. In brief though the backend routing is now about as simple as it gets (essentially dynamic backend selection by returning accessor methods as key lookups in a dict, and then blindly passing in a `hash` record value as input to the returned function which then "magically" returns the data.) The implementation details aren't really your concern when operating at the level this would need to work on,  but I'll just say that its rather efficient, and we should have no issues with multiple threads (or even processes) reading at the same time (even if they actually need to read from the same file on disk).

hhsecond:
@rlizzo Awesome, make sense. However, I would like to make a very simple (at least pytorch) data loader for the demo I have on 15th at PyCon Thailand. Then we can probably be iterating over it when we get more time for this. Although a utility in the core would be the best, I can write my own snippet example and tell the audience that something like this is coming into the core soon.

rlizzo:
Yeah if u can just write it as a user defined script on ur local system that would be best for now. Feel free to post it in a gist so we can use that as a jumping off point in the future.

 I don't want to limit you too much or sound too dictatorial, but it's important to me that we keep the code logically ordered (while maintaining ease of use). This is complicated enough as is, I won't accept bloat (I already want to kill some redundancies in the `records/queries.py` and `records/heads.py` and I think some of the functions in `records/parsing.py` are either unreachable or redundant right now...) Hope you can understand. We'll work together to make sure that these Bolton utilities can do what they need while respecting the structure of the core

hhsecond:
Make sense. Closing this for now!

PR #74, opened 2019-06-01
New logic for __len__

hhsecond:
## Motivation and Context
Logic that finds the length of dataset and metadata made better.

## Description
We fetch the count key first and then fetch the value of it instead of looping through the data samples or metadata

## Types of changes

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:
- [ ] My code follows the code style of this project.
- [ ] I have read the **[CONTRIBUTING](../CONTRIBUTING.rst)** document.
- [ ] I have signed (or will sign when prompted) the tensorwork CLA.

### DIFF
--- a/src/hangar/dataset.py
+++ b/src/hangar/dataset.py
@@ -128,7 +128,15 @@ class DatasetDataReader(object):
         int
             number of samples the dataset contains
         '''
-        return len(self._Query.dataset_data_names(self._dsetn))
+        if not self._is_conman:
+            self._dataTxn = self._TxnRegister.begin_reader_txn(self._dataenv)
+        try:
+            DatasetRecordCountKey = parsing.dataset_record_count_db_key_from_raw_key(self._dsetn)
+            dset_len = int(self._dataTxn.get(DatasetRecordCountKey))
+        finally:
+            if not self._is_conman:
+                self._dataTxn = self._TxnRegister.abort_reader_txn(self._dataenv)
+        return dset_len
 
     def __contains__(self, key):
         '''Determine if a key is a valid sample name in the dataset
--- a/src/hangar/metadata.py
+++ b/src/hangar/metadata.py
@@ -51,16 +51,16 @@ class MetadataReader(object):
         self._Query = RecordQuery(self._dataenv)
 
         self._is_writeable: bool = False
-        self.__is_conman: bool = False
+        self._is_conman: bool = False
 
     def __enter__(self):
-        self.__is_conman = True
+        self._is_conman = True
         self._labelTxn = TxnRegister().begin_reader_txn(self._labelenv)
         self._dataTxn = TxnRegister().begin_reader_txn(self._dataenv)
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        self.__is_conman = False
+        self._is_conman = False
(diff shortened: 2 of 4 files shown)


src/hangar/metadata.py
```diff
@@ -72,7 +72,15 @@ def __len__(self):
         int
             number of metadata key/value pairs.
         '''
-        return len(self._Query.metadata_names())
+        if not self.__is_conman:
```
hhsecond: Two things to note here. `self.__is_conman` was the old name we followed. I noticed you had changed it to `self._is_conman`.

src/hangar/dataset.py
```diff
@@ -128,7 +128,15 @@ def __len__(self):
         int
             number of samples the dataset contains
         '''
-        return len(self._Query.dataset_data_names(self._dsetn))
+        if not self._is_conman:
+            self._dataTxn = self._TxnRegister.begin_reader_txn(self._dataenv)
+        try:
+            DatasetRecordCountKey = parsing.dataset_record_count_db_key_from_raw_key(self._dsetn)
+            dset_len = int(self._dataTxn.get(DatasetRecordCountKey))
```
rlizzo: The top level dataset method should make no assumptions about the structure of the records returened from a `get` call. Instead we send it to the appropriate methods in the `parsing` module which will decode and return the data which is expected.  Instead, the implementation should read like this:

```
try:
    dsetCountKey = parsing.dataset_record_count_db_key_from_raw_key(self._dsetn)
    dsetCountVal = self._dataTxn.get(dsetCountKey, default='0'.encode())
    dset_count = parsing.dataset_record_count_raw_val_from_db_val(dsetCountVal)
finally:
   ...
```

There's actually a bug in the straight up `casting` implementation due to the fact that the string is not decoded. This might have failed in some corner cases...

Also, it's not stated anywhere, but if you look through the `dataset` or `metadata` files, you'll see that the following convensions are generally objserved for variable names. 

```
    camelCase = raw (binary) encoded data which is used to `get` / `set` keys/values in the db
    underscore (ie. `foo_val`) = some native python type or data structure (typically used for inputs/outputs).
```

If you find any cases where this convention isn't observed, please open a PR! (There very well may be some procedures still around which were from initial prototype of this tool before it was pushed to github and I actually established a bit more standard conventions/style)

src/hangar/metadata.py
```diff
@@ -72,7 +72,15 @@ def __len__(self):
...
-        return len(self._Query.metadata_names())
+        if not self._is_conman:
+            self._dataTxn = TxnRegister().begin_reader_txn(self._dataenv)
+        try:
+            metadataCountKey = parsing.metadata_count_db_key()
+            metadata_count = self._dataTxn.get(metadataCountKey)
+        finally:
+            if not self._is_conman:
+                self._dataTxn = TxnRegister().abort_reader_txn(self._dataenv)
+        return int(metadata_count) if metadata_count is not None else 0
```
rlizzo: lmdb transactions have the ability to set a default value if the lookup fails. This ternary check is unnecessary, and could have been eliminated with:

```
metaCountVal = self._dataTxn.get(metaCountKey, default='0'.encode())
```

hhsecond:
@rlizzo These are super helpful. Thanks a tonne. Closing this PR in favor of #70

Issue #56, opened 2019-05-23
[BUG REPORT] Test Cases Needed for New Merge/Diff Logic and API


rlizzo:
**Describe the bug**

Test cases are needed for the changes introduced in #32. 

**Severity**

Select an option:
- [ ] Data Corruption / Loss of Any Kind
- [ ] Unexpected Behavior, Exceptions or Error Thrown
- [ ] Performance Bottleneck

N/A

**To Reproduce**
Steps to reproduce the behavior, minimal example code preffered:

Run the tests

**Expected behavior**

More thorough testing.

**Screenshots**

N/A

**Desktop (please complete the following information):**

 - OS: any
 - Python: any
 - Hangar: 0.0.0 -> 0.1 rc

**Additional context**

See #32 for changes introduced

rlizzo:
Hi @hhsecond, can I assign this to you? I know you were starting to dig into the `diff` and `merger` internals the other day, and since this is stable (even considering the changes introduced in #70), its a good project which we won't duplicate effor on.

hhsecond:
@rlizzo Sure, I'll take it up.

rlizzo:
Fixed in #76

PR #76, opened 2019-06-10
Testing new diff logic

hhsecond:
## Motivation and Context
Testing diff logic. Addressing #56 

## Description
Describe your changes in detail:

## Types of changes

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:

### DIFF
--- a/tests/test_diff.py
+++ b/tests/test_diff.py
@@ -1,0 +1,89 @@
+import pytest
+import numpy as np
+
+
+@pytest.fixture()
+def repo_1_br_no_conf(repo):
+
+    dummyData = np.arange(50)
+
+    co1 = repo.checkout(write=True, branch_name='master')
+    co1.datasets.init_dataset(name='dummy', prototype=dummyData, samples_are_named=True)
+    for idx in range(10):
+        dummyData[:] = idx
+        co1.datasets['dummy'][str(idx)] = dummyData
+    co1.metadata['hello'] = 'world'
+    co1.commit('first commit adding dummy data and hello meta')
+    co1.close()
+
+    repo.create_branch('testbranch')
+    co2 = repo.checkout(write=True, branch_name='testbranch')
+    for idx in range(10, 20):
+        dummyData[:] = idx
+        co2.datasets['dummy'][str(idx)] = dummyData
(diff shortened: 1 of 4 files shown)


tests/test_diff.py
```diff
@@ -0,0 +1,118 @@
...
+        co = repo.checkout(branch_name='master')
+        diffdata = co.diff.branch('testbranch')
+        conflict_dict = diffdata[1]
+        assert conflict_dict['conflict_found'] is False
+        diffs1 = diffdata[0]
+
+        co = repo.checkout(branch_name='testbranch')
+        diffdata = co.diff.branch('master')
+        diffs2 = diffdata[0]
+
```
hhsecond: @rlizzo I have a question in the above block. As per my understanding, Diff created at line#91 & line#97 should have same data but flipped between `master` and `dev` (btw: this terminology will be super confusing for us in the future. We should name it as source/destination or something IMO). Is that right?

hhsecond:
@rlizzo I am sort of stuck with one logic here which I have mentioned. Once you could clarify that, I am good to finish up the other tests I have in mind.

tests/test_diff.py
rlizzo: So you are correct, since this is a Fast-Forward merge (as shown below). However, it should generally hold true even for 3-way merges (I think)

image

However, there is a bug which would not display the information correctly from the diff (which is corrected in #77). Try rebasing this on top of master and let me know if that makes it easier to understand? 

After the fix this is was shows: 

image

and 

image

hhsecond: Gotcha! thanks

hhsecond: Okey, thinking about it with a git analogy raises more questions. Perhaps we could talk about it over the call.

rlizzo: absolutely!

rlizzo:
Sorry for the delay @hhsecond, i'll sort through this tonight and get back to you soon!

hhsecond:
@rlizzo Yes, sorry for not getting back to this so far. Will finish it today or tomorrow.

rlizzo:
I've rebased this and am merging it now!

PR #91, opened 2019-07-16
Dataloaders for PyTorch & Tensorflow

hhsecond:
## Motivation and Context
PyTorch DataLoader for loading data from hangar directly into PyTorch

#13 

## Types of changes
- [ ] Documentation update

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:


rlizzo:
This is going to be great sherin! What's its status? Let me know what I can do since this and tensorflow loaders are going to become a priority for me in the next week or so.

hhsecond:
@rlizzo So the number of lines of code is very minimal here. I was planning to add IterableDataSet as well but that's not in the PyTorch release yet. I had to remove that. Now what's pending is adding tests, and make sure this line doesn't break `torch.from_numpy(self.hangar_dataset[key])`. Are you planning to work on TF Loader. I have it started already on my board (no code yet, so go ahead if you have something in your mind already)

rlizzo:
I haven't started anything yet a, and probably won't get to it for another
week so go ahead and submit that one as well. Does this work as expected in
it's current state?

hhsecond:
Yes, it should work as expected. Here is an example
```
from hangar import Repository
from hangar.dataloaders import TorchLoader
repo = Repository('.')
co = repo.checkout()
dset = co.datasets['cifar']
loader = TorchLoader(dset, batch_size=16)
for data in loader:
    print(data.shape)
```

elistevens:
This doesn't seem like it handles the case where the user has multiple columns of data they want to have loaded (here, "column" is the same concept as what's called a hangar dataset, and what I've been calling a Tensorset elsewhere).

I'd expect to see the TorchDataset taking a list of hangar datasets, so that the loader can provide (for example) the input and output tensors.

Am I missing something?

hhsecond:
@elistevens No, you are not. I just did not see how difficult (if not impossible) that is going to be when `shuffle` is ON (I was thinking multiple dataloaders for multiple tensorset, I know, I know, stupid me)

src/hangar/dataloaders/torch_loader.py
```diff
@@ -0,0 +1,97 @@
...
+        key = self._data_names[index]
+        if self._number_of_datasets == 1:
+            return torch.from_numpy(self.hangar_datasets[0][key])
+        out = []
+        for dset in self.hangar_datasets:
+            out.append(torch.from_numpy(dset[key]))
+        return out
+
+
+class TorchLoader(DataLoader):
```
elistevens: I would remove this class unless there are plans to add functionality to it. As it stands, there is no advantage to the user for using it.

src/hangar/dataloaders/torch_loader.py
```diff
@@ -0,0 +1,97 @@
...
+        self.hangar_datasets = hangar_datasets
+        self._data_names = tuple(self.hangar_datasets[0].keys())
+
+    def __len__(self):
+        return self._dataset_len
+
+    def __getitem__(self, index):
+        key = self._data_names[index]
+        if self._number_of_datasets == 1:
+            return torch.from_numpy(self.hangar_datasets[0][key])
```
elistevens: I would like the return type of the function to be consistent. Having to re-wrap items in a list for consistent downstream handling is annoying.

Also, in 1.1 if a dataset returns a named tuple, the data loader will preserve that tuple type as the return value. I'd like `__init__` to pick up a `tuple_cls=tuple` parameter, and have this function `return self._tuple_cls(out)`.

rlizzo:
Hey @hhsecond, I was able to fix the TravisCI build (the new extra package requirements were not set as dependencies of the top level `testenv` environment). Not sure what's going on with Appveyor right now though, seems like nothing should have changed. I suspect it has something to do with a cached package / build we have somewhere. I'll take a look tomorrow. I'll be sending the rest of my review comments in now.

hhsecond:
@rlizzo Awesome, Thanks

src/hangar/dataloaders/torch_loader.py
hhsecond: I disagree. It lets the user import only one loader and there is no need for doing the whole Dataset creation call. For those who need more control, we can make the `hangar.dataloaders.TorchDataSet` available. I agree with the fact that it doesn't do a lot but I see it as a convenience. After all, it doesn't harm.
That being said, I don't have any sentiment for keeping that, so if everyone thinks that should be gone, I am OK with that too.

src/hangar/dataloaders/torch_loader.py
hhsecond: `_tuple_cls` sounds brilliant and never knew about named tuple thing. What do you think about making 1.1 the lowest version we support?

hhsecond: Never knew new named tuple thing, thanks for the pointer. So `tuple_cls` seemed brilliant but that actually creates more problems. For example, `tuple` expect a single argument which can be a list/tuple while named tuple expect n number of arguments where n is the number of fields. We'll have to do another check for that which I think is weirder. 
Also, the rationale behind returning non-consistent return type is although it is inconsistent in different cases, it is consistent with the input type. If the input type is single `hangar.Dataset` output is a single batch. If the input is a list of `hangar.Dataset` output is a list of batches. Also, regardless of tuple/list torch DataSet returns, Dataloader always returns a list.
So I think we can accept two types of inputs, 
1. `hangar.Dataset`
2. `list` / `tuple` of `hangar.Dataset`

and if `tuple_cls` (default to `None`) is defined then without doing any check we'll return `tuple_cls(out)` otherwise return the output as it is

What do you think?

elistevens: If I can get a one-element tuple in the one-dataset case by setting `tuple_cls` to `tuple` then I'm happy.

You can check for `hasattr(self._tuple_cls, '_fields')` to detect the named tuple case.  :)

hhsecond: That's the thing, PyTorch DataLoaders convert that tuple to a one-element list. So you break the consistency of return data type there itself.

rlizzo:
I'm going to warn you that it's not safe to import the DatasetDataReader (or really any of the dataset or metadata class methods) here, or quite frankly, anywhere. It's even avoided in the hangar core so that we don't place those classes on the import path. The reason is because we need to ensure that only hangar internally has a strong reference keeping the object live. When it's no longer safe to access the object (the checkouts been closed) we can guarantee that once we delete that strong reference, garbage collector will clean it up, end of the weakref proxies to the objects are invalidated. 

I think you're just using it as a type annotation here, and if that's actually the case just use a forward ref to the class which hasn't actually been imported.

hhsecond:
@rlizzo Make sense but yes, I am using that only for an instance check. I could use some other check for sure.

rlizzo:
Yeah. Honestly I don't know if it's even necessary to do a check. Duck typing should be good enough for the `DatasetDataReader` class.

rlizzo:
This needs to be rebased

src/hangar/dataloaders/torch_loader.py
hhsecond: Dataset now returns a list, tuple or a namedtuple. Nothing else

src/hangar/dataloaders/torch_loader.py
hhsecond: Okay, @elistevens I am going to apologize for the disagreement. With the introduction of new keys, I couldn't find a way to keep the TorchLoader clean which is worthy enough. So removing it.

src/hangar/dataloaders/torch_loader.py
hhsecond: I am thinking of returning only namedtuple with hangar dataset names as field names (if the user doesn't pass a `tuple_cls`) What do you think @elistevens

hhsecond:
@rlizzo I have changed the `testenv`'s extras back to `all` which is required for CIs. Let me know if you have other plans

src/hangar/dataloaders/torch_loader.py
elistevens: Oh, I like that idea a lot.

hhsecond:
@rlizzo I am kinda stuck on the sorting order here. Don't you think we should keep the insertion order of data just like python dicts instead of lexical?

hhsecond:
So apparently, `pip install torch` doesn't work well in Windows. We might need to add `test_requirements.txt` file if we want to run pytorch tests on windows for now.

src/hangar/dataloaders/torch_loader.py
```diff
@@ -0,0 +1,118 @@
...
+    >>> dset = co.datasets['dummy_dset']
+    >>> torch_dset = TorchDataset(dset, index_range=slice(1, 100))
+    >>> loader = DataLoader(torch_dset, batch_size=16)
+    >>> for batch in loader:
+    ...     train_model(batch)
+
+    """
+
+    def __init__(self, hangar_datasets, keys=None, index_range=None, field_names=None):
+        if hasattr(hangar_datasets, 'get') and hasattr(hangar_datasets, 'name'):
```
elistevens: why not "if not isinstance(hangar_datasets, list):`?

src/hangar/dataloaders/torch_loader.py
```diff
@@ -0,0 +1,118 @@
...
+            dset_names.append(dset.name)
+        self.hangar_datasets = hangar_datasets
+
+        if field_names:
+            if len(field_names) != len(hangar_datasets):
+                raise RuntimeError("length mismatch: `field_names` and `hangar_datasets` must "
+                                   "have same number of elements")
+            self.nt_wrapper = namedtuple('BatchDataset', field_names=field_names)
+        else:
+            self.nt_wrapper = namedtuple('BatchDataset', field_names=dset_names, rename=True)
```
elistevens: Could this be named `'BatchTuple'` instead? It's not a Dataset (technically it won't be a batch either, until after the data loader, so `'SampleTuple'` but since the intended use case is to use a data loader, so it will be correct at that point).

src/hangar/dataloaders/torch_loader.py
hhsecond: No good reason. Just felt safe with checking only two attributes of `Dataset` which I needed to use downstream

src/hangar/dataloaders/torch_loader.py
hhsecond: That's fine

hhsecond:
@elistevens I have added TF dataloaders and made few API changes for the whole dataloader module to keep the consistency. It's not the final version. I am doing more testing & documentation but would be great if you can take a look

hhsecond:
@rlizzo can you give it a go when you get time

rlizzo:
I'm looking it over now. How do suggest we mark it as `"experimental"`? At the moment we don't have a real process in place for that.

hhsecond:
@rlizzo I was thinking about mentioning that in the doc, docstring and where ever we speak or even be a warning that gets printed to the terminal on the invocation.

src/hangar/__init__.py
```diff
@@ -4,3 +4,15 @@
 from .repository import Repository
 
 __all__ = ['Repository', 'serve']
+
+try:
+    from .dataloaders import make_tf_dataset
```
elistevens: I would skip the indirection; exposing the same function in multiple different modules isn't particularly pythonic. I'd just pick one place to get the functions from; it doesn't really matter where that place is (`hangar.make_xxx_dataset` or `hangar.dataloaders.make_xxx_dataset` or `hangar.dataloaders.tfloader.make_tf_dataset` all work). I think that I lean towards the top level, but that's a weak preference.

src/hangar/dataloaders/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+    from .tfloader import make_tf_dataset
+    __all__.append('make_tf_dataset')
+except RuntimeError:
+    pass
+
+try:
+    from .torchloader import make_torch_dataset
+    __all__.append('make_torch_dataset')
+except RuntimeError:
+    pass
```
elistevens: Given my comments above, this file could be entirely empty.

src/hangar/dataloaders/common.py
```diff
@@ -0,0 +1,96 @@
...
+        dset_keys, dset_lens = {}, []
+        for dataset in hangar_datasets:
+            kset = set(dataset.keys())
+            dset_keys[dataset.name] = kset
+            dset_lens.append(len(kset))
+
+        if not np.allclose(dset_lens):
+            warnings.warn('Datasets do not contain equal num samples', UserWarning)
+
+        if (keys is not None) and (not isinstance(keys, (list, tuple, set))):
```
elistevens: What is this doing? It's not clear what wrapping `keys` in a tuple is supposed to do.

src/hangar/dataloaders/common.py
```diff
@@ -0,0 +1,96 @@
...
+        new `index_range` is setup.
+        """
+        if keys:
+            if not isinstance(keys, (list, tuple)):
+                raise TypeError(f"`keys` must be a list/tuple, not {type(keys)}")
+            sample_names = keys
+        elif index_range:
+            if not isinstance(index_range, slice):
+                raise TypeError(f"`index_range` must be a slice, not {type(keys)}")
+            sample_names = tuple(self.dataset_array[0].keys())[index_range]
```
elistevens: You're using tuple here as essentially a `frozenlist` right? The immutability is the point? I'd include a comment saying as much, since tuple to me means "fixed collection of disparate data items" not "immutable list".

elistevens:
Overall, looks great. Very happy with the API!

src/hangar/dataloaders/common.py
hhsecond: Allowing users to pass single key (A string)

hhsecond:
@rlizzo I can't think of a better way to move dataloader utilities to the top level without the hack I have now. Although I am more lenient towards moving it under `dataloader`, it makes sense to go with majority

tox.ini
```diff
@@ -1,13 +1,36 @@
...
+    pytest-xdist
+
+[mldeps]
+deps =
+    {[standarddeps]deps}
+    tensorflow
+    https://download.pytorch.org/whl/cpu/torch-1.1.0-cp36-cp36m-win_amd64.whl ; sys_platform == 'win32'
+    https://download.pytorch.org/whl/cpu/torchvision-0.3.0-cp36-cp36m-win_amd64.whl ; sys_platform == 'win32'
+    torch ; sys_platform == 'darwin'
+    torchvision ; sys_platform == 'darwin'
```
hhsecond: @rlizzo Why are we doing `torchvision` here. I don't think we need that anywhere?

rlizzo:
@hhsecond, FYI that I just force pushed a rebase on top of the master (including changes in #104).  I'm updating the `datasets` -> `arraysets` now and will push another fix shortly.

src/hangar/dataloaders/common.py
```diff
@@ -0,0 +1,107 @@
...
+            self._allowed_samples = tuple(keys)
+        elif index_range:
+            if not isinstance(index_range, slice):
+                raise TypeError(f"`index_range` must be a slice, not {type(keys)}")
+            self._allowed_samples = tuple(sample_names)[index_range]
+        else:
+            self._allowed_samples = tuple(sample_names)
+
+        self.dataset_array = hangar_datasets
+        self.dataset_names = dataset_names
```
rlizzo: This whole init section is really pretty complicated. what exactally are we trying to accomplish here?

src/hangar/dataloaders/common.py
```diff
@@ -0,0 +1,108 @@
...
+            self._allowed_samples = tuple(keys)
+        elif index_range:
+            if not isinstance(index_range, slice):
+                raise TypeError(f"`index_range` must be a slice, not {type(keys)}")
+            self._allowed_samples = tuple(sample_names)[index_range]
+        else:
+            self._allowed_samples = tuple(sample_names)
+
+        self.arrayset_array = arraysets
+        self.arrayset_names = arrayset_names
```
rlizzo: This entire init section is pretty ugly. Can't this be simplified?

rlizzo:
Hey @hhsecond I've made all the updates and have done a bit of cleanup work here as well. Can you add some more testcases on larger sized arraysets, as well as a few which verify the output is what is expected? Once you're done let me know and I'd say we are good to merge!

rlizzo:
Also, if you wanted to add some notes in the documentation that could be helpful!

hhsecond:
@rlizzo I can make the existing test cases working on bigger arraysets than `(5,7)` but do we really need that. For testing the shapes, we are already doing it with the current test cases. Also, I have a bit more clean up in mind which makes `GroupedAset` abstraction bit more pythonic. This is something I would like to push along with the few examples I am working on currently. So with the next release, I can push
- Advanced examples
- Changes to the internals of dataloaders if required to include the corner cases
- Documentation

Let's merge what we already have then?

rlizzo:
Ok. I made a few changes which prevent writer checkout arrraysets from being used.

For missing tests, I'd say testing on different backends manually (both hdf5 and numpy memmap), testing to make sure no oddness occurs if there are arraysets with different backends in the same loader. 

Handling the case were there there are remote references as the data (which show up as keys, but not as real data). Handling large amounts of data to see what the memory profile of many keys is

Issue #81, opened 2019-07-01
[BUG REPORT] Commit inside context manager throws RuntimeError


hhsecond:
**Describe the bug**
If we try to commit inside the context manager (before `__exit__()`), hangar throws RuntimeError saying `No changes made in the staging area. Cannot commit.`. We should allow the user to do commits inside the context manager IMO but probably with a warning about the performance hit

**Severity**

Select an option:
- [ ] Data Corruption / Loss of Any Kind
- [x] Unexpected Behavior, Exceptions or Error Thrown
- [ ] Performance Bottleneck

**To Reproduce**
```python
import numpy as np

from hangar import Repository
repo = Repository(path='myhangarrepo')
repo.init(user_name='Sherin Thomas', user_email='sherin@gmail.com', remove_old=True)

# generate data
data = []
for i in range(1000):
    data.append(np.random.rand(28, 28))
data = np.array(data)

co = repo.checkout(write=True)
data_dset = co.datasets.init_dataset('mnist_data', prototype=data[0])
co.commit('datasets init')
co.close()
co = repo.checkout(write=True)
data_dset = co.datasets['mnist_data']

with data_dset:
    for i in range(len(data)):
        sample_name = str(i)
        data_dset[sample_name] = data[i]
        co.commit('dataset curation: stage 1')  # this throws error
co.close()
```

**Expected behavior**
It should not break the program instead raise a warning about the performance hit

hhsecond:
@lantiga I remember telling you this won't break the program instead raise an issue. I was wrong. It raise a `RuntimeError` and crashes

lantiga:
@hhsecond ok, thanks for bringing this up. One question (also cc @rlizzo): is there any other fundamental issue (current or future) that makes it important to not commit within the context, other than performance?

I think throwing an exception is not wrong in itself. The intention for the context is to group operations that you can then commit, so enforcing no commits reinforces that intention. @hhsecond do you see a use case where this is impractical?

One last question: @hhsecond when you say that
> It raises a `RuntimeError` and crashes
you mean that if you catch the exception we're good or it leaves the repo in an unworkable state?

BTW: we should also be able to iterate over the dataset. Maybe also having a `named_iterator` would be useful, so we're not inducing that `sample_name = str(i)`, which I don't find kosher.

hhsecond:
So here are my thoughts.
- I don't think there are any other fundamental issues. Correct me if I am wrong @rlizzo 
[...]

PR #94, opened 2019-07-18
Commit inside context manager

hhsecond:
## Motivation and Context
Commits inside a context manager is not handled in the master

#81 

## Description
When commit is being triggered

- It calls context manager teardown function which flushes the data to disk
- then executes the commit operation
- then call context manager setup function

## Types of changes
- [ ] Documentation update

- [ ] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:

### DIFF
--- a/src/hangar/checkout.py
+++ b/src/hangar/checkout.py
@@ -3,6 +3,8 @@ import logging
 import weakref
 from uuid import uuid4
 from os.path import join as pjoin
+from contextlib import suppress
+from functools import partial
 
 import lmdb
 
@@ -96,13 +98,13 @@ class ReaderCheckout(object):
         PermissionError
             if the checkout was previously close
         '''
-        try:
-            assert hasattr(self, '_metadata')
-            assert hasattr(self, '_datasets')
-        except AssertionError:
-            err = f'Unable to operate on past checkout objects which have been '\
-                  f'closed. No operation occurred. Please use a new checkout.'
-            raise PermissionError(err) from None
+        p_hasattr = partial(hasattr, self)
+        if not all(map(p_hasattr, ['_metadata', '_datasets', '_differ'])):
+            e = PermissionError(
+                f'Unable to operate on past checkout objects which have been '
+                f'closed. No operation occurred. Please use a new checkout.')
+            logger.error(e, exc_info=False)
+            raise e from None
 
     @property
     def datasets(self) -> Datasets:
@@ -110,8 +112,8 @@ class ReaderCheckout(object):
 
         .. seealso::
 
-            The class :class:`hangar.dataset.Datasets` contains all methods accessible
-            by this property accessor
+            The class :class:`hangar.dataset.Datasets` contains all methods
+            accessible by this property accessor
 
         Returns
         -------
@@ -130,8 +132,8 @@ class ReaderCheckout(object):
 
         .. seealso::
 
-            The class :class:`hangar.metadata.MetadataReader` contains all methods accessible
-            by this property accessor
+            The class :class:`hangar.metadata.MetadataReader` contains all methods
+            accessible by this property accessor
 
         Returns
         -------
@@ -185,11 +187,35 @@ class ReaderCheckout(object):
         multiple simultaneous read checkouts.
         '''
         self.__verify_checkout_alive()
-        self._datasets._close()
+        with suppress(AttributeError):
+            self._datasets._close()
+
+        for dsetn in (self._datasets._datasets.keys()):
+            for attr in list(self._datasets._datasets[dsetn].__dir__()):
+                with suppress(AttributeError, TypeError):
+                    delattr(self._datasets._datasets[dsetn], attr)
+
+        for attr in list(self._datasets.__dir__()):
+            with suppress(AttributeError, TypeError):
+                # adding `_self_` addresses `WeakrefProxy` wrapped by `ObjectProxy`
+                delattr(self._datasets, f'_self_{attr}')
+
+        for attr in list(self._metadata.__dir__()):
+            with suppress(AttributeError, TypeError):
+                # adding `_self_` addresses `WeakrefProxy` wrapped by `ObjectProxy`
+                delattr(self._metadata, f'_self_{attr}')
 
         del self._datasets
         del self._metadata
         del self._differ
+        del self._commit_hash
+        del self._repo_path
+        del self._labelenv
+        del self._dataenv
+        del self._hashenv
+        del self._branchenv
+        del self._refenv
+        return
 
 
 # --------------- Write enabled checkout ---------------------------------------
@@ -312,8 +338,8 @@ class WriterCheckout(object):
 
         .. seealso::
 
-            The class :class:`hangar.metadata.MetadataWriter` contains all methods accessible
-            by this property accessor
+            The class :class:`hangar.metadata.MetadataWriter` contains all methods
+            accessible by this property accessor
 
         Returns
         -------
@@ -332,8 +358,8 @@ class WriterCheckout(object):
 
         .. seealso::
 
-            The class :class:`hangar.diff.WriterUserDiff` contains all methods accessible
-            by this property accessor
+            The class :class:`hangar.diff.WriterUserDiff` contains all methods
+            accessible by this property accessor
 
         Returns
         -------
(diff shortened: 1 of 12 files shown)


src/hangar/dataset.py
```diff
@@ -429,6 +421,26 @@ def _backend(self) -> str:
         '''
         return self._dflt_backend
 
+    def _cm_setup(self):
```
rlizzo: I'm very confused here. Why are we moving the methods to handle context manager setup and teardown out of the standard python methods? 

Am I missing something? Or are we just redirecting the call? If so. Consider that this breaks the manual attributes set on the weakref object proxy which is returned to the user upon access. (a `functools partial` instance applied with all arguments filled but unevaluated to call the CM protocol of a weakref proxy via manually set attributes on a reference python doesn't normally think is `hashable`)  

At best you might describe it as an elegant hack which is thoroughly tested. At worst it's a mindbend 2 lines of code applied librally to the User facing classes which took me a week to write, will give most python experts a significant pause when they see it, and will just be ignored by everyone else. 

I don't want to be a pain and make others pay for a difficult design decision, but if we don't have a strong need to break with standard python CM conventions, it'll be easier to use them all around.

src/hangar/checkout.py
```diff
@@ -532,6 +532,13 @@ def commit(self, commit_message):
         '''
         self.__acquire_writer_lock()
         logger.info(f'Commit operation requested with message: {commit_message}')
+        metadataconman = self._metadata._is_conman
+        if metadataconman:
```
rlizzo: You might need to do some additional checks on the transaction register here. While normally each teardown of the metadata and dataset class instances will either abort or commit pending reader transactions from their threads, you need to consider the state of all of them and not just one. 

The transaction register is just a reference counting scheme which shares transaction handles between multiple concurrent requesters 
 When you're operating within a context manager at the checkout level you'll have to care  that each outstanding transaction handle has actually been closed. I don't know if it's a concern in practice, but this is an area where you want to make sure that all writes have been actually flushed to disk

Does that make sense?

src/hangar/dataset.py
```diff
@@ -366,20 +366,12 @@ def __init__(self, stagehashenv, default_schema_backend, *args, **kwargs):
 
     def __enter__(self):
         self._is_conman = True
-        self._hashTxn = self._TxnRegister.begin_writer_txn(self._hashenv)
-        self._dataTxn = self._TxnRegister.begin_writer_txn(self._dataenv)
-        self._stageHashTxn = self._TxnRegister.begin_writer_txn(self._stagehashenv)
-        for k in self._fs.keys():
-            self._fs[k].__enter__()
+        self._cm_setup()
```
hhsecond: @rlizzo I have a gut feeling that you are not going to like this. But I am not really a big fan of calling `__enter__` & `__exit__` from outside :(

src/hangar/dataset.py
hhsecond: So none of this is well tested, I have just made it up (which I should not have I have verified it well I guess). So the reason for taking it out of standard methods is, I can rely on `self._is_conman` while re-enabling it after the commit. Otherwise, my `conman` becomes False on `__exit__` and I'll have to keep a reference separately. I mean, that's OK and probably there could be other methods but I just found this easy. Thanks for the pointer, I'll check it out.

rlizzo:
Hey @hhsecond this needs to be rebased ontop of master before we evaluate the PR

src/hangar/checkout.py
hhsecond: @rlizzo I can't think of an instance this is going to break. So leaving it as it is. Let me know if it rings a bell for you.

rlizzo:
Ok @hhsecond some interesting stuff ended up happening here. Long story short your additions worked very well. you were right in thinking that they wansn't a code-path which would lead to us needing to check the `TxnRegister`. 

However...

During my own tests I notices that the following behavior would occur when using context mangers. 

```python
>>> co = repo.checkout(write=True)
    # same behavior occurred for ``write=False``

>>> with co.metadata as m:
...     m['hello'] = 'world'
...     cmt1 = co.commit('here is the second commit')   
...     with co.datasets['dummy'] as d:
...         d['1'] = np.arange(10)
...         cmt2 = co.commit('again commit')
    
>>> co.close()
OK

>>> repr(co)
Unable to operate on past checkout objects which have been closed.  No operation occurred. Please use a new checkout.
---------------------------------------------------------------------------
PermissionError                           Traceback (most recent call last)
~/projects/tensorwerk/hangar/hangar-py/src/hangar/checkout.py in _repr_pretty_(self, p, cycle)
    268         '''pretty repr for printing in jupyter notebooks
    269         '''
--> 270         self.__acquire_writer_lock()
    271         res = f'\n Hangar {self.__class__.__name__}\
    272                 \n     Writer       : True\

~/projects/tensorwerk/hangar/hangar-py/src/hangar/checkout.py in __acquire_writer_lock(self)
    440                   f'closed. No operation occurred. Please use a new checkout.'
    441             logger.error(err, exc_info=0)
--> 442             raise PermissionError(err) from None
    443 
    444         try:
PermissionError: Unable to operate on past checkout objects which have been closed. No operation occurred. Please use a new checkout.
-----------------------------------------------------------------------------
# Note: That error is what we want to happen

# This, however, is not!
>>> repr(d)
Hangar DatasetDataWriter                 
    Dataset Name             : dummy                
    Schema Hash              : 051a8e928804                
    Variable Shape           : False                
    (max) Shape              : (50,)                
    Datatype                 : <class 'numpy.int64'>                
    Named Samples            : True                
    Access Mode              : a                
    Number of Samples        : 27                
    Partial Remote Data Refs : False

# we even had full access to the data files on disk as they were being
# kept alive by the references. 
>>> d['1']
np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

>>> repr(m)
Hangar Metadata                
    Writeable: False                
    Number of Keys: 1

>>> m['hello']
'world'
```

Long story short, there was an issue because context managers need to return a true instance of `self` in their `__enter__()` methods. By utilizing the fact that we are returning a handle to the `weakref.Proxy` through an `ObjectProxy`, if we do a much more thorough cleanup of the objects down the stack, everything will actually work out. 

That's what i've included in the rest of this PR. Take a look and I'm good to merge when you are!

src/hangar/checkout.py
rlizzo: See my comment https://github.com/tensorwerk/hangar-py/pull/94#issuecomment-513514480 below. I'm good with this!

src/hangar/checkout.py
```diff
@@ -185,11 +187,35 @@ def close(self) -> None:
         multiple simultaneous read checkouts.
         '''
         self.__verify_checkout_alive()
-        self._datasets._close()
+        with suppress(AttributeError):
+            self._datasets._close()
+
+        for dsetn in (self._datasets._datasets.keys()):
+            for attr in list(self._datasets._datasets[dsetn].__dir__()):
+                with suppress(AttributeError, TypeError):
```
hhsecond: Minor nit: can't we use a single `with` statement for all three loops?

src/hangar/checkout.py
```diff
@@ -451,18 +475,12 @@ def __acquire_writer_lock(self):
         try:
             heads.acquire_writer_lock(self._branchenv, self._writer_lock)
         except PermissionError as e:
-            try:
+            with suppress(AttributeError):
```
hhsecond: same as above; may use the same `with` statement? and perhaps delete other attributes as well

src/hangar/diagnostics/ecosystem.py
```diff
@@ -64,13 +64,13 @@ def get_optional_info() -> dict:
 
 def version_of_package(pkg):
     """ Try a variety of common ways to get the version of a package """
-    with ignoring(AttributeError):
+    with suppress(AttributeError):
```
hhsecond: single `with`

src/hangar/checkout.py
rlizzo: actually no, since if a `with suppress(FooException):` block encounters an Exception, it will exit the `with` block and continue execution at the next line outside the block.

src/hangar/checkout.py
rlizzo: see https://github.com/tensorwerk/hangar-py/pull/94#discussion_r310359814

Also I don't want to delete other attributes since the object should still be "`introspect`-able". We just need to delete any attributes which could access data on disk.

src/hangar/diagnostics/ecosystem.py
rlizzo: see https://github.com/tensorwerk/hangar-py/pull/94#discussion_r310359814

PR #102, opened 2019-08-05
Rename datasets to datacell

rlizzo:

The name `Datasets` no longer fit the appropriate description as a container of tensor/array data.

In order to make this clearer, `datasets` has been replaced by the term `datacells`. Since this more accuratly describes the ability for a single sample in a dataset to be made up of individual pieces spread across `datacells`. 

The new explantation is in full [HERE](https://github.com/rlizzo/hangar-py/blob/f30b1495242143d2109cce10065a79c50b6ddf19/docs/concepts.rst#how-hangar-thinks-about-data), but for the sake of brevity, this diagram illustrates the "unique" relationship that a dataset has to `samples` and `datacells`:

```
   A Dataset is thought of as containing Samples, but is actually defined by
    Datacells, which store parts of fully defined Samples in structures
       common across the full aggregation of Samples in the Dataset

   _____________________________________
         S1     |    S2    |     S3     |  <------------------------|
   --------------------------------------                           |
       image    |  image   |   image    |  <- Datacell 1  <--|      |
     filename   | filename |  filename  |  <- Datacell 2  <--|-- Dataset
       label    |  label   |   label    |  <- Datacell 3  <--|
     annotation |    -     | annotation |  <- Datacell 4  <--|


   If a sample does not have a piece of data, lack of info in the Datacell
         makes no difference in any way to the larger picture.
```

## Types of changes
- [ ] Documentation update

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:

Please give this a review @lantiga @hhsecond


elistevens:
I think that data cell sounds like a single entry (`S3.image`), not the entire column.

The spreadsheet metaphor almost works, except that it's a key to tensor mapping, not an indexed list of tensors, and that the mapping can be sparse. If you are going to use the spreadsheet metaphor, this should be a "data column".

In that case, transpose your example, so that image, file name, etc. Are column headings and each sample is a row.

Then you say that a data set is just a view over your columns, that gives you sample tuples based on the cross product of keys and columns. Hangar doesn't store or track the data set, just the underlying columns.

rlizzo:
Thanks for the comments @elistevens! Do you think that calling it a "Datacolumn" instead would evoke too many comparisons to a columnar DB? The term is a bit clearer, but that would be a very misleading interpretation to take. 

Part of the reason I ended up deciding on `datacell` is because in this context, "cell" has no obvious meaning / prior definition which we need to either fit the mental model to, or redefine. 

Another term I seriously considered was `datagroup`. Is that any better?

elistevens:
Data group is exceedingly generic, but it works, I think.

lantiga:
Hey, chiming in!

I think both datacell and datacolumn are very obscure. They aren't wrong, mind you, but we need to try to make it easy for users to grasp the basic concepts and this doesn't help IMHO.

lantiga:
I'm sure I missed out, but why did we drop `tensorset`?

rlizzo:
**If we can agree to `datagroup` then I think that's the way to go.**

For `tensorset`, it's a combination of the fact that (formally speaking) a tensor and an array are not actually identical, me not liking the sound of the name, and that outside of the AI community (general data analytics/processing) people more commonly talk about `arrays` or `lists` and not `tensors` (at least in my experience). it's best to just be generic and say `data = numeric bytes + shape + dtype (regardless of what your community calls it)`

In addition, `set` is the wrong suffix to add. If we don't rely on the semantic meaning of `dataset`, then it's best not to imply that it's a `set-like` data structure (it's more of a `map` really). Generic allows us to define it in our own way, not in the way people think about some specific term.

rlizzo:
Also Eli, you are totally right about the impression that the diagram gives off! 

that totally needs to be redone.

lantiga:
I think datagroup will confuse the heck out of users, honestly I don't know what that means, it could mean so many things.

rlizzo:
But that's what gives us the fledibility to say "yeah, this is a bit of a generic name, but that's because we don't care about the specifics of your data. Give each `datagroup` a name which means something to you and a compatible shape/dtype and go on with your life"...

also, it won't hurt that this is really the only way to interact with data on disk, so we have every chance to explain how it should be interpreted.

rlizzo:
I'm at a loss for other alternatives. We just need to pick something and be consistent.

lantiga:
That's the issue IMHO, our data are tensors (or arrays, just pick one). We don't support tabular data, or other kinds of data, so why be generic?  We need to be precise also in what we don't provide.

I don't like datagroup because a group reminds me of HDF5 groups, I've never seen `group` used as a collection of something, and it's not just data but tensors (or arrays).

This is a tensor collection, no?

rlizzo:
The more this gets discussed on the user group and over the phone the more debate there is and the less we agree. If this wasn't something which is one of the most typed out API bits in all of hangar, we could probably eventually get to something longer which is more descriptive.

Since that's not the case, we are left with the one thing we can all semi get behind (no one hates it atleast): `tensorset` or `arrayset`. 

I'm choosing `arrayset`. 

PS: I don't want this to come off as sounding too harsh, or make it appear that I am ungrateful for all of everyones input. I really do appreciate it, and the discussions alone have been massively interesting (and will make for some great documentation). This is just something which we could go back and forth for weeks on and still not make a decision (and actually, we kindof already have been...). 0.2 is ready to be pushed aside from these last few details. A decision just has to be made.

rlizzo:
closing this PR due to #104 which superseeds changes here.

PR #110, opened 2019-08-19
Making Dataloaders stable

hhsecond:
Fixing few issues with Dataloaders

## Types of changes
- [ ] Documentation update

- [ ] Ready for review
- [x] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:

### DIFF
--- a/docs/Hangar-Tutorial-003.ipynb
+++ b/docs/Hangar-Tutorial-003.ipynb
@@ -336,17 +336,17 @@
     "  \n",
     "    - Is that the same commit I'm trying to push? if yes, abort\n",
     "    \n",
-    "    - Is that a commit that I dont have? If yes, someone else has updated that branch, abort\n",
+    "    - Is that a commit I dont have? If yes, someone else has updated that branch, abort\n",
     "    \n",
     "- Here's the commit digests which are parents of my branches head, which commits are you missing?\n",
     "\n",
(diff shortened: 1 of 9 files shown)


rlizzo:
@hhsecond FYI I rebased this to make the tests pass with hotfox in #113

hhsecond:
@rlizzo this PR must be in a good state to merge now. We need to take a decision on what hangar should do when the data is not available locally. Throwing an error might not be a good idea for a long-running training loop. Maybe we could rase a warning about some data not being available locally right when the `make_torch/tf_dataset` call. What do you think?

rlizzo:
> @rlizzo this PR must be in a good state to merge now. We need to take a decision on what hangar should do when the data is not available locally. Throwing an error might not be a good idea for a long-running training loop. Maybe we could rase a warning about some data not being available locally, on the `make_torch/tf_dataset` call. What do you think?

Hey @hhsecond I reviewed this (made a few changes) and am about to merge! Nice Job!

What I decided to do was to check for remote reference data up front when the `make_torch_dataset` or `make_tf_dataset` methods are first called. the `allowed_samples` in the `GroupedAsets` class are then the intersection of `common_keys`, `local_sample_keys`, and (if specified) `user_keys`. If not-local samples are specified, we raise a `ValueError` up front instead of letting the user progress to the training loop. I think this is an OK solution for the time being. I've modified and added tests to ensure this behavior is correct.

Issue #79, opened 2019-06-14
[FEATURE REQUEST] Multi-tensor samples from complex data sources


elistevens:
**Is your feature request related to a problem? Please describe.**
I want to know how to accomplish the following. It doesn't need to require zero effort on the user's part, but there needs to be a clear best-hangar-practices path to a workable setup.

Take a source data format that is complex (e.g. DICOM, JPEG) and infeasible to reconstitute bit-exact from the tensor+metadata form. Each instance of the raw data produces a sample that consists of 2 (or more) tensors; an image tensor and a 1D tensor that encodes things like lat/long or age/sex/etc. (to be concatenated with the output of the convolutional layers prior to the fully connected layers). To be clear, this is intended to be an illustrative example, not a concrete use case.

Per my reading of the docs, right now these two tensors wouldn't qualify as being in the same hangar dataset (it's not clear if that's problematic or not).

Let's express the above conversion as:

`f_v1(raw) -> (t1, t2)`

Users will need to:

- Update the conversion function to `f_v2` and repopulate `t1` and `t2`.
- Update the conversion function to `f_v3` which outputs `(t1, t2, t3)`.
- Update the raw data for a sample and repopulate `t1` and `t2`.
- Be handed the pair of `t1` and `t2` for training/validation (including when training is randomized).
- Retrieve the raw data given IDs/tags/metadata included with the training sample (for use in an external viewer, manual investigation, etc.).

**Describe the solution you'd like**
I think that changing the definition of a sample to be a tuple of binary blobs plus a tuple of tensors plus metadata would work, but I haven't considered the potential impacts from that kind of change. Seems potentially large.

**Describe alternatives you've considered**
Another option would be to have separate datasets for `t1` and `t2` and combine them manually, plus manage the binary blobs separately. That seems like a lot of infra work, and might be at risk of having drift between the samples themselves, and with the blobs.

**Additional context**
I suspect that I want/expect Hangar to solve a larger slice of the problem than it's intended to, but it's not clear at first glance what the intended approach would be for more complicated setups like the above.

lantiga:
Hey @elistevens, thanks for your patience, we should have gotten to this earlier.

So, that’s right, a Dataset in Hangar maps to a single schema (a single collection of tensors). A Repo, on the other hand, can have more than one Datasets, so, for instance, to store data where two different pieces of info sit alongside each other you’d have a Repo with two Datasets. The samples in the dataset would have the same sample `name`.

[...]

PR #115, opened 2019-08-20
API Redesign

rlizzo:

To simplify user interface with arraysets and provide some concept of a `dataset` as a view across arraysets.

***NOTE:** This initial PR is a proof of concept only, and will require extensive discussion before the final design is agreed upon*

related to #79 and many conversations on the Hangar Users Slack Channel

Added `CheckoutIndexer` class which is inhereted in `ReaderCheckout` and `WriterCheckout` to enable the following API. (originally proposed by @lantiga and @elistevens)

```python
dset = repo.checkout(write=True)
# get an arrayset of the dataset (i.e. a "column" of the dataset?)
aset = dset['foo']

# get a specific array from 'foo' (returns a named tuple)
arr = dset['foo', '1']
# set it too
dset['foo', '1'] = arr

# get data from dset (returns a named tuple)
subarr = dset['foo', '1']
# and set into it
dset['foo', '1'] = subarr + 1

# get a sample of a dataset across 'foo' and 'bar' (returns a named tuple)
sample = dset[('foo', 'bar'), '1']

# get a sample of all arraysets in the checkout (returns a named tuple)
sample = dset[:, '1']
sample = dset[..., '1']

# get multiple samples
sample_ids = ['1', '2', '3']
batch = dset[('foo', 'bar'), sample_ids]
batch = dset[:, sample_ids]
batch = dset[..., sample_ids]
```

## Types of changes
- [ ] Documentation update

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:


hhsecond:
@rlizzo @lantiga I found it more intuitive to call 
```python
sample = dset[('foo', 'bar')][1]
```
It might add another layer over arraysets (something like `GroupedAset` from dataloaders) to handle the subindex. What do you think?

rlizzo:
> @rlizzo @lantiga I found it more intuitive to call
> 
> ```python
> sample = dset[('foo', 'bar')][1]
> ```
> 
> It might add another layer over arraysets (something like `GroupedAset` from dataloaders) to handle the subindex. What do you think?

I actually think that's slightly confusing, since to specify multiple `arraysets` or `samples` you have to write parentheses `(foo, bar, baz)` inside of the bracket style notation `[(foo, bar, baz)]`. However, if we removed the parentheses, specifying sample would look like:

```python
# case 1
samples = dset['foo', 'bar']['1', '2']

# case 2
samples = dset[:]['1']

# case 3
samples = dset[...]['1', '2']

# case 4
samples = dset['foo']['1']
```

While `case 2` and `case 4` look OK, `1` and `3` look a bit odd to me (thinking as a numpy user). Having this nested sort of access structure doesn't really give off the same impression that the returned results are a view of `sample names` over `arrayset names` (to me at least). 

With `samples = dset[['foo', 'bar'], 1]`, I know what i'm going to get, and it's less to type. @elistevens, would love to hear your input on this one as well?

### Another Idea

We actually may want to take some minor inspiration from the [xarray](http://xarray.pydata.org/en/stable/#) project, as it is the de-facto standard for working with labelled ndim arrays, or as they say:

> xarray (formerly xray) is an open source project and Python package that makes working with labelled multi-dimensional arrays simple, efficient, and fun!

While we wouldn't want a "one-to-one" mapping of API and access conventions, their approach to [positional indexing](http://xarray.pydata.org/en/stable/indexing.html#positional-indexing) is essentially the same as the proposed `samples = dset[['foo', 'bar'], '1']` method (substitution int indexes for named indexes). [vectorized indexing](http://xarray.pydata.org/en/stable/indexing.html#vectorized-indexing)  is the view of multiple samples over multiple arraysets `samples = dset[['foo', 'bar'], ['1', '2']]`, and [multi-level indexing](http://xarray.pydata.org/en/stable/indexing.html#multi-level-indexing) could be rather adaptable to us as well. Maybe: 

```python
# vectorized indexing
samples = dset[['foo', ['1', '2']], ['bar', ['3', '4']]`?
```

We could also allow "dict style" indexing: `samples = dset[{'foo': '1', 'bar': '1'}]`, though having too many options isn't always desirable. 

I would specifically avoid their "Pandas style" named indexing (using `.iloc(foo)` and `.sel(foo)`). However, that's only since it's a lot of complexity to add to what is meant to be a "convenience" method. 

### One more Issue

`xarray` has also have decided what happens to [missing coordinate labels](http://xarray.pydata.org/en/stable/indexing.html#missing-coordinate-labels), which is an issue I was about to raise. We need to decide what happens in the following case

```python
# checkout has arraysets 'foo', 'bar', 'baz'.
# 'foo' has samples '1', '2'
# 'bar' has samples '1', '2'
# 'baz' has samples '1'    <-- missing '2'

# Case 1 (expected output)
>>> res = dset[:, '1']
>>> print(res)
('foo'=np.array([1]), 'bar'=np.array([1, 1]), 'baz'=np.array([1, 1, 1])

# Case 2A
>>> res = dset[:, '2']
KeyError('2' not in arrayset 'baz')

# Case 2B
>>> res = dset[:, '2']
>>> print(res)
('foo'=np.array([2]), 'bar'=np.array([2, 2]), 'baz'=None)

# Case 2C
>>> res = dset[:, '2']
>>> print(res)
('foo'=np.array([2]), 'bar'=np.array([2, 2]))
```

I was leaning toward suggesting `Case 2C` before I read the `xarray` spec, but it turns out that is the exact behavior they implement as well. Is there any reason we shouldn't adopt that pattern? (regardless of other conventions we decide on)?

elistevens:
I think that the talk of how to layer syntactic sugar over the underlying operations is happening way to early in the design process. Trying to decide how many square brackets you need to get the data out shouldn't happen until the release after you've got verbose methods implemented and released.

Also, keep in mind that 99% of your users are _never_ going to use this API. Data will be written, committed, and then they'll call the API to convert it to a Dataset they can feed into their deep learning framework of choice. I'm not saying that because I think that it's okay to design it poorly, just that it's not a great use of time right now.

There's an unfortunate trend in a lot of the more research/data science tooling to paper over data problems, which I think runs entirely counter to the hangar "data integrity is paramount" ethos. If I ask for something, but some of the keys are missing, _I should get an exception_. (FWIW, I feel the same way about missing dimensions, but that's a different PR)

I do think that it's okay to have a keyword arg that enables silently discarding rows that don't meet the column constraints, but I should have to enable it explicitly (and I think that needing it points to deeper problems in the data pipeline).

lantiga:
Hey, great job @rlizzo! Various coments:

I agree with @rlizzo's comment on double brackets; I also started with @hhsecond's approach but switched to multi-indexing later on as it was simpler to grasp. When I say that, I'm referring to the fact that our users post 0.3 coming to Hangar will be primarily exposed this API to accessing / writing datasets, as it doesn't require further terminology or object hierarchy than just what you checkout, and it suffices for the majority of the use cases.
I imagine that what they will need to learn is that you checkout something and that you just index into it with named indexes.

@rlizzo it's great that we can get consistent with XArrray. I personally think `2C` is the right call (I was originally thinking about returning `None` for that key), it makes the behavior about a dataset as an indexed subset of a checkout simpler (you can iterate over keys with an ellipse). We can add another method (like `get`) or an option in the square brackets a-la `R` (yak!) to make it `throw`.

@elistevens it's a good warning re: being too soon, but I don't agree with this. One thing we know is that we need to simplify Hangar's API to make users grow. I do think this is the API that 99% of users should use, because it won't require them to go deep into Hangar's design details to make it useful. The current API will be for advanced users in my head.

rlizzo:
@hhsecond, can you take a look at this and let me know what you think about testing for these new access methods? Not sure if we should be repeating many of the low level tests or just writing new ones for the high level API?

src/hangar/checkout.py
```diff
@@ -127,6 +134,236 @@ def __verify_checkout_alive(self):
...
+
+        """
+        try:
+            tmpconman = not self._is_conman
+            if tmpconman:
+                self.__verify_checkout_alive()
+                self.__enter__()
+
+            # Arrayset Parsing
+            if (arraysets is (Ellipsis, slice)):
```
hhsecond: `in` instead of `is`?

rlizzo:
Thanks for starting. I'm gonna say go ahead and give it a look now! I'll come back to the rest of the tests once you give it a once over.

hhsecond:
@rlizzo We need something to replace `co.arraysets.keys()` we had in the older API to get the names of all arraysets

hhsecond:
Do we have a plan to enable users to loop over the dataset where `dataset = repo.checkout()`

src/hangar/checkout.py
```diff
@@ -127,6 +134,236 @@ def __verify_checkout_alive(self):
...
+        *  Slice syntax for arrayset field cannot specify `start`, `stop`, or
+           `step` fields, it is soley a shortcut syntax for 'get all arraysets' in
+           the ``:`` or ``slice(None)`` form
+
+        .. seealso:
+
+            :meth:`~hangar.checkout.ReaderCheckout.get`
+
+        """
+        try:
```
hhsecond: IMO, we should offload all of these tasks to `get` method and `__getitem__` will just do the `index` parsing and call `get`. This way we can avoid double checking and make sure our semantics never is broken.

src/hangar/checkout.py
```diff
@@ -127,6 +134,236 @@ def __verify_checkout_alive(self):
...
+                            raise e
+                        arr = None
+                    aset_samples.append(arr)
+                if nAsets == 1:
+                    asetsSamplesData = aset_samples
+                    if nSamples == 1:
+                        asetsSamplesData = asetsSamplesData[0]
+                    break
+                asetsSamplesData.append(aset_samples)
+            else:  # N.B. for-else conditional (ie. 'no break')
```
hhsecond: nice!!

src/hangar/checkout.py
```diff
@@ -127,6 +134,236 @@ def __verify_checkout_alive(self):
...
+                tmp = map(ArraysetData._make, zip(*asetsSamplesData))
+                asetsSamplesData = list(tmp)
```
hhsecond: thought: list comprehension instead of map and then list?

src/hangar/checkout.py
```diff
@@ -127,6 +134,236 @@ def __verify_checkout_alive(self):
...
+                        arr = None
+                    aset_samples.append(arr)
+                if nAsets == 1:
+                    asetsSamplesData = aset_samples
+                    if nSamples == 1:
+                        asetsSamplesData = asetsSamplesData[0]
+                    break
+                asetsSamplesData.append(aset_samples)
+            else:  # N.B. for-else conditional (ie. 'no break')
+                tmp = map(ArraysetData._make, zip(*asetsSamplesData))
```
hhsecond: thought: list comprehension instead of map and then list?

src/hangar/checkout.py
rlizzo: Nice catch!

Actually it's a bit of a weird case here. `Ellipses` is an object literal and needs a actual `is` check (like comparing if something is `None`), however, the comparison to `slice` needs a `isinstance` check... 

Should be `if (arraysets is Ellipsis) or isinstance(arraysets, slice):` fixing it now!

src/hangar/checkout.py
rlizzo: I'll have to check on this one. `map` generates a loop which executes in the compiled `C-python` layer. I tend to use it whenever possible since it offers a decent speedup for long running loops. Not sure if we would see much of a difference here though if we used standard list comp. I'll let ya know.

rlizzo:
> @rlizzo We need something to replace `co.arraysets.keys()` we had in the older API to get the names of all arraysets

Why? The standard API isn't going anywhere, and is really important to work with anything other than a pre-built repo/dataset. Is there some reason we shouldn't just keep `co.arraysets.keys()`?

rlizzo:
> Do we have a plan to enable users to loop over the dataset where `dataset = repo.checkout()`

It's not on my radar right now, and I'm not sure what it would even return?? `arraysets keys`, `metadata`? Values? 

I'm actually very wary of actually treating the `Checkout` object as a dataset in much more of a capacity than it is now. It actually has some super important methods (`commit`, `diff`, `merge`, `reset_staging_area`, etc..) let alone providing access to the standard `Arrayset` and `Metadata` class methods. 

from my point of view, these changes are mearly a conveience for a common use case users will face, they aren't changing the fundumental nature of what the `Checkout` object actually does. I just wouldn't want to send mixed signals about what each layer is responsible for. 

Does that make sense? Maybe I'm seeing this wrong though. What's your point of view?

src/hangar/checkout.py
```diff
@@ -127,6 +134,236 @@ def __verify_checkout_alive(self):
...
+            Single arrayset, multiple samples array data for each sample is
+            returned in same order sample keys are recieved.
+
+        List[NamedTuple[*np.ndarray]]
+            Multiple arraysets, multiple samples. Each arrayset's name is used
+            as a field in the NamedTuple elements, each NamedTuple contains
+            arrays stored in each arrayset via a common sample key. Each sample
+            key is returned values as an individual element in the
+            List. The sample order is returned in the same order it wasw recieved.
+
```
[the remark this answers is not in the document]
rlizzo: It might be nice to have that in some cases, but since there isn't anyway to set some modifiable `flag` during dict style access it would have to be permenant. Two levels of indexing to pull elements which could be stored sequentially in a list would drive me nuts...

rlizzo:
Documentation updated and all tests pass with good coverage. Ready to be merged!

hhsecond:
> > @rlizzo We need something to replace `co.arraysets.keys()` we had in the older API to get the names of all arraysets
> 
> Why? The standard API isn't going anywhere, and is really important to work with anything other than a pre-built repo/dataset. Is there some reason we shouldn't just keep `co.arraysets.keys()`?

It is just more intuitive to get all the relevant information from the high-level API itself, IMO

hhsecond:
> > Do we have a plan to enable users to loop over the dataset where `dataset = repo.checkout()`
> 
> It's not on my radar right now, and I'm not sure what it would even return?? `arraysets keys`, `metadata`? Values?
> 
> I'm actually very wary of actually treating the `Checkout` object as a dataset in much more of a capacity than it is now. It actually has some super important methods (`commit`, `diff`, `merge`, `reset_staging_area`, etc..) let alone providing access to the standard `Arrayset` and `Metadata` class methods.
> 
> from my point of view, these changes are mearly a conveience for a common use case users will face, they aren't changing the fundumental nature of what the `Checkout` object actually does. I just wouldn't want to send mixed signals about what each layer is responsible for.
> 
> Does that make sense? Maybe I'm seeing this wrong though. What's your point of view?

Yep, make sense. Thanks!

src/hangar/checkout.py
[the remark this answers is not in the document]
hhsecond: No, what I am saying is, It is already returning a NamedTuple in the case `dset[('foo', 'bar', 'baz'), '1']` which is not documented in the `Returns` field. Or Am I missing something here?

src/hangar/checkout.py
```diff
@@ -350,10 +636,488 @@ def __repr__(self):
...
+            nSamples = len(sampleNames)
+
+            # Verify asets
+            if (nAsets > 1) and (nSamples > 1):
+                raise SyntaxError(
+                    'Not allowed to specify BOTH multiple samples AND multiple'
+                    'arraysets in `set` operation in current Hangar implementation')
+
+            elif (nAsets == 1) and (nSamples == 1):
+                for aset in asets:
```
hhsecond: small nit: can't we do it without loops. It looks ugly to do the loop since you are doing the len check.

rlizzo: fair enough. fixed!

PR #130, opened 2019-10-09
Remotes Refactor

rlizzo:
#### _Why is this change required? What problem does it solve?:_

This is a long overdue refactor and general update to the remote client/server logic. Among a lot of cleanup and simplification, this update removes two unnecessary dependencies (`msgpack` and `pyyaml`), improves remote test coverage, and offers a number of performance gains. 

This is a WIP.

#### _Describe your changes in detail

Will provide when nearing a more final state.

## Types of changes
What types of changes does your code introduce? Put an `x` in all the boxes that apply:
- [ ] Documentation update
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

Is this PR ready for review, or a work in progress?
- [ ] Ready for review
- [x] Work in progress

## How Has This Been Tested?
Put an `x` in the boxes that apply:
- [ ] Current tests cover modifications made
- [x] New tests have been added to the test suite
- [x] Modifications were made to existing tests to support these changes
- [ ] Tests may be needed, but they are not included when the PR was proposed
- [ ] I don't know. Help!

## Checklist:
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **[CONTRIBUTING](../CONTRIBUTING.rst)** document.
- [x] I have signed (or will sign when prompted) the tensorwork CLA.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #134, opened 2019-10-11
Plugins revamp

hhsecond:
## Motivation and Context
Revamping plugin system to make at actually pluggable for different data types etc.

## Description
- Introducing new `io` module. This will help users to use the import/export functionality of plugins through the program if they don't wan't to interact with the low-level hangar APIs
- We potentially can move other modules like `dataset` inside `io` module
- Test cases are work in progress
- Docs are work in progress

## Types of changes
- [x] Documentation update

- [ ] Ready for review
- [x] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:


hhsecond:
@rlizzo There will be few more minute changes but it's in a good shape now to review especially because of the introduction of `io` module

rlizzo:
Ok. I'd like to merge the changes in #130 before this, since they did a big overhaul of the test suit to simplify it's fixture setup (improving speeds in most cases)

hhsecond:
Make sense. I'll finish my review soon

rlizzo:
Hey @hhsecond lets connect about these changes on a call. I don't think I agree with moving to `io` directory to top level of `src/hangar`. Will get into the details over a vid chat.

hhsecond:
Sure thing. I was reluctant too about it initially but then realized few reasons why should we do it. Let's talk

src/hangar/cli/utils.py
```diff
@@ -0,0 +1,10 @@
+
+def parse_custom_arguments(click_args):
```
rlizzo: What does this function do? I'm confused. Why does it need it's own module?

src/hangar/external/manage_plugins.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+def _scan_plugins():
+    """
+    Scan the plugins directory for .ini files and parse them
+    to autofill storage
+    """
+    pd = os.path.dirname(__file__)
+    config_files = glob(os.path.join(pd, '_plugins', '*.ini'))
+
```
rlizzo: So how does someone add a plugin? The files need to go into the hangar/external/_plugins directory?

src/hangar/external/manage_plugins.py
```diff
@@ -0,0 +1,219 @@
...
+                    if tmp_prv:
+                        tmp_prv.append(name)
+                    else:
+                        tmp_acc[provides] = [name]
+                else:
+                    plugin_details_store[accepts] = {provides: [name]}
+        plugin_module_name[name] = os.path.basename(filename)[:-4]
+
+
+def _read_preferred_plugins_config():
```
rlizzo: I don't understand why this is needed? It's messy, and how would it be extended by other plugins?

src/hangar/external/manage_plugins.py
```diff
@@ -0,0 +1,219 @@
...
+    Populate preferred plugin dict that maps file formats to plugins and methods. This
+    is used to infer which plugin to use at runtime based on file format
+    """
+    parser = configparser.ConfigParser()
+    config_path = os.path.join(os.path.dirname(__file__), 'preferred_plugins.ini')
+    parser.read(config_path)
+    return parser
+
+
+def _write_preferred_plugins_config(parser: configparser.ConfigParser):
```
rlizzo: It's not ok to dynamically write source files to a library. This would get way to complex to debug in the future.

src/hangar/external/manage_plugins.py
```diff
@@ -0,0 +1,219 @@
...
+        meta_data[opt] = parser.get(name, opt)
+    return name, meta_data
+
+
+def _default_name_sampler(fpath):
+    return os.path.splitext(os.path.basename(fpath))[0]
+
+
+def reset_plugins():
+    global _preferred_plugins
```
rlizzo: Does this actually need to be a global? or should we be using a `Plugin` class here?

src/hangar/external/manage_plugins.py
```diff
@@ -0,0 +1,219 @@
...
+    with open(config_path, 'w') as f:
+        parser.write(f)
+
+
+def _import(plugin: str, method: str) -> Callable:
+    if plugin in _plugin_store:
+        loaded_plugin = _plugin_store[plugin]
+    else:
+        modname = plugin_module_name[plugin]
+        loaded_plugin = __import__(f'hangar.external._plugins.{modname}', fromlist=[modname])
```
rlizzo: use `importlib.import_module` instead of `__import__` method.

src/hangar/external/preferred_plugins.ini
```diff
@@ -0,0 +1,54 @@
+[jpg]
```
rlizzo: Not sure why this file exists... can't we get all this info from what's provided in each plugins `.ini` file?

tests/test_external.py
```diff
@@ -0,0 +1,40 @@
...
+    yield None
+    external.reset_plugins()
+
+
+@pytest.fixture()
+def reset_preferred_plugins():
+    preferred_plugins = copy.deepcopy(external.manage_plugins._preferred_plugins)
+    yield None
+    external.manage_plugins._preferred_plugins = preferred_plugins
+    external.manage_plugins._write_preferred_plugins_config(preferred_plugins)
```
rlizzo: This is a prime candidate for monkeypatching. Preferably with the `context manager` form to auto-reset past values. 

https://docs.pytest.org/en/latest/monkeypatch.html#global-patch-example-preventing-requests-from-remote-operations

src/hangar/external/manage_plugins.py
elistevens: I'm only skimming comments, but if plugins don't support something like "install this plugin from pypi; activate by calling `somepluginpkg.activate()`" then the plugin system is DOA for use cases like mine (which require that we get wheels and check the SHA before we use them in our build system).

src/hangar/cli/utils.py
hhsecond: Keeping `utils` there for any future utils that we could have (in fact there might be existing logic in cli.py which could be moved to utils). 
`parse_custom_arguments` is the function that takes the unknown arguments from the command line (https://github.com/hhsecond/hangar-py/blob/9d70a533f92a3c0efba1cec7d0b49ff345b3cd16/src/hangar/cli/cli.py#L364) which is plugin-specific. For example, if the user wants to pass the dimension of images to be while doing the import, he could pass `--height` and `--width` if plugin is designed to work on those. Does this make sense?

src/hangar/external/manage_plugins.py
hhsecond: OK, I am sorry I did not touch this part. This was the existing logic and we probably need to move it to user home directory or something. And yes @elistevens, the idea is to copy the required files on `pip install` or something like `plugin.activate`. The plan is to make a cookie-cutter package/repo that does this and plugin developers need to work only on the methods such as `load`
Now that when I think about it, the whole copy thing could be avoided by having each plugin provides some convenient way to let hangar knows what functions are they exposing and what file format they could work with

src/hangar/external/manage_plugins.py
hhsecond: It is just to read the `preferred plugin configuration` that has user's preferences. Why would it be extended to other plugins?

src/hangar/external/manage_plugins.py
hhsecond: Right, moving it out.

src/hangar/external/manage_plugins.py
hhsecond: hmm, we could. Although I don't see any troubles by using global in this particular use case. It was essentially a clean up of what we had in the existing `manage_plugin.py`. I'll make a PluginManager class

src/hangar/external/manage_plugins.py
hhsecond: Fair enough

src/hangar/external/preferred_plugins.ini
hhsecond: These are user preferences. We need to make user preferences stateful and hence this file. We don't get this information from plugin's `ini`

tests/test_external.py
hhsecond: yep, will do that

src/hangar/external/manage_plugins.py
elistevens: Sorry, I was unclear: my usage absolutely will not allow files in the virtual env to be modified at runtime.

Have a function to register a plugin that takes a class with some callback methods or something like that.

elistevens:
What's the actual use case for plugins? I'm worried it's a lot of complexity and API surface area for benefit that isn't super clear from the outside.

src/hangar/external/manage_plugins.py
hhsecond: Okey, We can do without the copy. I did not do much on the plugin registration and discovery part and was trying to make it actually pluggable for all data types (earlier the plugin system was explicitly coded for image data).

I am considering using package metadata for plugin discovery now which avoids the need for the copy and will be registered on `pip install` itself
https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata. So the flow is going to be like this

- cookie cutter repo will have file structure and setup.py which does what is required for plugin discovery
- Developers who create plugin should create a class and should inherit from `hangar.external.BasePlugin`. This parent class will have the required functions for letting the hangar knows what functions are exposed through the plugin and what are the file formats it could work with
- A plugin can expose `load`, `save`, `show` and `board_show`. Where `load` is for loading the data from the disk and return numpy array, `save` takes a numpy array from hangar repo and save that to the provided path, `show` takes numpy array from hangar repo and show like what matplotlib show does, `board_show` takes numpy array from hangar repo and returns in a form which is understood by javascript to display (this is for the hangar board)

Am I clear enough?

hhsecond:
@elistevens comment here https://github.com/tensorwerk/hangar-py/pull/134#discussion_r336724026 should give you more info. Lemme know what do you think

elistevens:
Sounds great!

src/hangar/cli/utils.py
rlizzo: yup! good idea for this one. Thanks!

src/hangar/external/manage_plugins.py
rlizzo: So me and sherin have discussed this, and we are going to use `entry_points` and a system similar to https://github.com/mitsuhiko/pluginbase to enable and activate extensions. This should solve the issues you mentioned @elistevens. Thanks for the feedback!

src/hangar/external/manage_plugins.py
rlizzo: Resolved due to discussion on vid call. Will rely on user specification or just default to first available parser for the format.

src/hangar/external/manage_plugins.py
rlizzo: Resolved due to discussion over vid chat. Will be handled by in memory structures to avoid this.

src/hangar/external/preferred_plugins.ini
rlizzo: Resolved from vid chat discussion. User plugins may implement `.ini` file (which we will parse) or dict mapping in same `ini` style defined in `entry_points` hook.

tests/test_cli.py
```diff
@@ -383,4 +381,155 @@ def test_start_server(managed_tmpdir):
...
+        assert np.allclose(d1, d2)
+
+        # with overwrite
+        res = runner.invoke(cli.import_data, [aset_name, fpath, '--overwrite'], obj=repo)
+        assert res.exit_code == 0
+        co = repo.checkout(write=True)
+        co.commit('added data')
+        d3 = co.arraysets[aset_name][fpath]
+        co.close()
+        assert not np.allclose(d1, d3)
```
rlizzo: Move all individual `"checks"` into their own tests. It's not apparent that these are individual test cases, and that the first two (should) fail, the third is the "garden path", and the last two test optional flags/behavior. 

setting up the repo as a fixture and using named tests would be much clearer here.

tests/test_cli.py
```diff
@@ -383,4 +381,155 @@ def test_start_server(managed_tmpdir):
...
+        print(fpath)
+
+    with monkeypatch.context() as m:
+        m.setattr(PluginManager, "_scan_plugins", monkeypatch_scan(['save'], ['ext'], 'save', save))
+        res = runner.invoke(
+            cli.export_data, [aset_name, '-o', str(tmp_path), '--sample', 'data', '--format', 'ext'], obj=repo)
+        assert res.exit_code == 0
+        assert save_msg in res.output
+
+        # invalid plugin
```
rlizzo: see previous comment on splitting up tests.

tests/test_cli.py
```diff
@@ -383,4 +381,155 @@ def test_start_server(managed_tmpdir):
...
+    with monkeypatch.context() as m:
+        m.setattr(PluginManager, "_scan_plugins", monkeypatch_scan(['show'], ['ext'], 'show', show))
+
+        res = runner.invoke(
+            cli.view_data, [aset_name, 'data', '--format', 'ext'], obj=repo)
+        assert res.exit_code == 0
+        assert show_msg in res.output
+
+        # with startpoint
+        res = runner.invoke(
```
rlizzo: Is there a "successful" run of this method here?

src/hangar/cli/cli.py
```diff
@@ -359,136 +360,172 @@ def server(overwrite, ip, port, timeout):
...
+    elif startpoint:
         base_commit = expand_short_commit_digest(repo._env.refenv, startpoint)
+    else:
+        branch_name = get_staging_branch_head(repo._env.branchenv)
+        base_commit = get_branch_head_commit(repo._env.branchenv, branch_name)
+    co = repo.checkout(commit=base_commit)
+
+    sample, stype = sample.split(':') if ':' in sample else (sample, '')
+    if stype == 'int' or stype == 'str':
+        sample = eval(stype)(sample)
```
rlizzo: don't use eval.

```suggestion
        sample = int(sample) if stype == 'int' else str(sample)
```

src/hangar/cli/cli.py
```diff
@@ -359,136 +360,172 @@ def server(overwrite, ip, port, timeout):
...
+    co = repo.checkout(commit=base_commit)
 
     try:
-        co = repo.checkout(write=False, commit=base_commit)
-        arrayset = co.arraysets[arrayset]
+        aset = co.arraysets.get(arrayset)
         if sample:
+            sample, stype = sample.split(':') if ':' in sample else (sample, '')
+            if stype == 'int' or stype == 'str':
+                sample = eval(stype)(sample)
```
rlizzo: don't use eval.

```suggestion
                sample = int(sample) if stype == 'int' else str(sample)
```

src/hangar/cli/cli.py
```diff
@@ -359,136 +360,172 @@ def server(overwrite, ip, port, timeout):
...
-                try:
-                    a[fn] = arr
-                except ValueError as e:
-                    click.echo(e)
+            files = [path]
+        with aset, click.progressbar(files) as filesBar:
+            for fpath in filesBar:
+                extension = os.path.splitext(fpath)[1].strip('.')
+                loaded = external.load(fpath, plugin=plugin, extension=extension, **kwargs)
+                if isinstance(loaded, types.GeneratorType):
```
rlizzo: If we are always passing in a single file path (and not a directory), why would this be a `GeneratorType`? wouldn't  "one file in" always result in "one piece of data out" ?

hhsecond: If you meant how are we handling the single file case, that's the `else`

src/hangar/cli/cli.py
elistevens: Better yet, have a `stype_to_type = {'int': int, 'str': str}` dict somewhere (maybe even somewhere plugins can add things to), and dispatch on that.

hhsecond: Done

src/hangar/cli/cli.py
hhsecond: Done

tests/test_cli.py
hhsecond: line #519

tests/test_cli.py
hhsecond: I was trying to save few seconds (setup time) here but it makes sense to keep it your way for readability

tests/test_cli.py
hhsecond: Done

src/hangar/external/_external.py
```diff
@@ -0,0 +1,135 @@
...
+        command line argument, all the unknown keyword arguments will be collected
+        as `plugin_kwargs`
+
+    Returns
+    -------
+    img_array : ndarray
+        ndarray returned from the given plugin
+
+    """
+    pm = PluginManager()
```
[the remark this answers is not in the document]
hhsecond: There will be a circular import but I can work around my way

src/hangar/cli/cli.py
rlizzo: Ahh ok, i was a bit confused here. It wasn't immediately clear that this was designed to handle situations where a file can output many different samples to store in an arrayset. Highlighting this with something like a CSV file (in a test or comment) would be helpful for future readers..

tests/test_cli.py
rlizzo: Ok. thanks!

src/hangar/cli/cli.py
hhsecond: Good point, Will do that

setup.py
```diff
@@ -21,13 +21,7 @@ def read(*names, **kwargs):
...
 
 extra_require = {
-    'dev': [
-        'grpcio_tools',
-        'mypy>=0.701',
-        'mypy-protobuf',
-        'pillow',
-        'matplotlib',
-    ],
+    'plugins': ['hangar_pil', 'hangar_matplotlib'],
```
rlizzo: These should not be requirements of the hangar project itself. If the user wants them, they should download them. If we do this, it means every installation of hangar will download PIL, Matplotlib, and all of their dependencies. It's a lot of downloading to do if you don't need it.

src/hangar/cli/cli.py
```diff
@@ -535,136 +536,172 @@ def server(overwrite, ip, port, timeout):
...
+    at times. It is possible to do that by separating the name and type by a
+    colon.
+
+    Example:
+
+       1. if the sample name is string of numeric 10 - ``10:str`` or ``10``
+
+       2. if the sample name is ``sample1`` - ``sample1:str1`` or ``sample1``
+
+       3. if the sample name is an int, let say 10 - ``10:int``
```
rlizzo: @lantiga, do you have any suggestions here?

this is the best interface we could come up with, since apparently every other "simple" identifier for a number or a string seems to have some significance to a bash shell before it gets parsed by python.  

Our preferred were: 

- `"3"` to identify a `string 3` (`"` and `'` are escaped) 
- `#3` for an `int 3`(`#` starts a comment)
- ``` `3` ``` for an `string 3` (``` ` ``` enclosed value is directly executed as a command)

src/hangar/cli/cli.py
```diff
@@ -535,136 +536,172 @@ def server(overwrite, ip, port, timeout):
...
-                arr = imread(fpth, plugin=plugin)
-                try:
-                    a[fn] = arr
-                except ValueError as e:
-                    click.echo(e)
+            files = [path]
+        with aset, click.progressbar(files) as filesBar:
+            for fpath in filesBar:
+                extension = os.path.splitext(fpath)[1].strip('.')
+                loaded = external.load(fpath, plugin=plugin, extension=extension, **kwargs)
```
rlizzo: Missed corner case: file with multiple extensions

`foo.png` -> `png`
`foo.tar.bz2` -> `bz2`

i'm fixing this so:

`foo.tar.bz2` -> `tar.bz2`

src/hangar/cli/cli.py
```diff
@@ -535,136 +536,172 @@ def server(overwrite, ip, port, timeout):
...
-        co = repo.checkout(write=False, commit=base_commit)
-        arrayset = co.arraysets[arrayset]
+        aset = co.arraysets.get(arrayset)
         if sample:
+            sample, stype = sample.split(':') if ':' in sample else (sample, '')
+            sample = int(sample) if stype == 'int' else str(sample)
             sampleNames = [sample]
         else:
-            sampleNames = list(arrayset.keys())
+            sampleNames = list(aset.keys())
```
rlizzo: @hhsecond This is some seriously buggy code. 

- If i specify a sample `1:str` and `1:int`, when converted to a file name, they would be the same. 
 - same thing for if I don't specify file names, and there are mixed `str`, `int` keys, they would just end up overwriting eachother. 

Don't know what you'd like to do here, but this is bad that the tests aren't thorough enough to catch this.

I'll let you decide on what the fix is and implement it (along with the tests to prove it)

src/hangar/cli/cli.py
elistevens: IMO: If it parses as an int, it's an int. 

If it starts with `str:` or `int:` then it's the thing indicated.

Otherwise, it's a string.

elistevens: Note that this means that all `int:` does is prevent fallback to string in the case where it doesn't parse cleanly. So `int:3` works, and `int:foo` causes the entire program to bomb out.

src/hangar/cli/cli.py
```diff
@@ -535,136 +536,172 @@ def server(overwrite, ip, port, timeout):
...
 
-        if format_:
-            format_ = format_.lstrip('.')
-        outP = os.path.expanduser(os.path.normpath(out))
+        extension = format_.lstrip('.') if format_ else None
+        outdir = os.path.normpath(os.path.expanduser(outdir))
+        if not os.path.isdir(outdir):
+            click.echo(f"Directory {outdir} does not exist")
+            return
```
rlizzo: you don't need to do any of this. I'm pushing up a change that shows the fix implemented in the `import_data` method, but briefly. 

- use the `click.Path` type in the option specification. for this method this would make click check the path exists, is a directory, is not a file, is writeable by the user, and is fully resolved (including expanding symlinks / hardlinks - which this code doesn't catch)
```suggestion
@click.option('-o', '--out', 'outdir',
                required=True,
                type=click.Path(exists=True, dir_okay=True, file_okay=False, writeable=True, resolve_path=True))
```

Also, we should be using `click.ClickException` instead of "echo and return". This will still except immediatly, but won't print the traceback by default, and makes it possible during testing to ensure that this particular case raised an error. I don't know if the tests cover this case now, but please make sure they do. See my changes in `import_data` for an example.

src/hangar/cli/cli.py
elistevens: Also, this is bikeshedding, but I much prefer the prefix form to the suffix.

Also, if it's possible for the value to have a colon in it, then I would require the `str:` prefix.

src/hangar/cli/cli.py
```diff
@@ -535,136 +536,172 @@ def server(overwrite, ip, port, timeout):
...
-                    else:
-                        outFP = os.path.join(outP, f'{sampleN}.{format_}')
-                else:
-                    outFP = os.path.join(outP, f'{sampleN}')
                 try:
                     data = aset[sampleN]
-                    imsave(outFP, data)
+                    external.save(data, outdir, sampleN, extension, plugin, **kwargs)
                 except KeyError as e:
                     click.echo(e)
```
rlizzo: Same thing here.  as use `click.Exception`

src/hangar/cli/cli.py
```diff
@@ -535,136 +536,172 @@ def server(overwrite, ip, port, timeout):
...
+        branch_name = get_staging_branch_head(repo._env.branchenv)
+        base_commit = get_branch_head_commit(repo._env.branchenv, branch_name)
+    co = repo.checkout(commit=base_commit)
+
+    sample, stype = sample.split(':') if ':' in sample else (sample, '')
+    sample = int(sample) if stype == 'int' else str(sample)
+
+    extension = format_.lstrip('.') if format_ else None
 
     try:
```
rlizzo: The repo was checked out way to early here, if the code between it and the `try` block every raises an error, the program would never get to the point where it would close.

src/hangar/cli/cli.py
```diff
@@ -535,136 +536,172 @@ def server(overwrite, ip, port, timeout):
...
-        arrayset = co.arraysets[arrayset]
+        aset = co.arraysets.get(arrayset)
         try:
-            data = arrayset[sample]
-            imshow(data, plugin=plugin)
-            show()
+            data = aset[sample]
+            external.show(data, plugin=plugin, extension=extension, **kwargs)
         except KeyError as e:
             click.echo(e)
```
rlizzo: Use `click.ClickException` (see previous messages)

src/hangar/cli/cli.py
```diff
@@ -535,136 +536,172 @@ def server(overwrite, ip, port, timeout):
...
+        base_commit = get_branch_head_commit(repo._env.branchenv, branch_name)
+    co = repo.checkout(commit=base_commit)
 
     try:
-        co = repo.checkout(write=False, commit=base_commit)
-        arrayset = co.arraysets[arrayset]
+        aset = co.arraysets.get(arrayset)
         if sample:
+            sample, stype = sample.split(':') if ':' in sample else (sample, '')
+            sample = int(sample) if stype == 'int' else str(sample)
```
rlizzo: Instead of doing this same parsing logic in every one of the plugin methods, you should be using a custom parameter type. That way you could just say 

```python
@click.option('-s', '--sample',
               type=utils.StrOrIntType, 
               default=None, 
                help='Sample name to export. You can optionally specify the type'
                      ' of sample name after the colon (ex. `54:str` or `54:int`)')
```

and each element in the sample list would be automatically parsed and converted by that function before it even hits our function here. 

Please read up on it: https://click.palletsprojects.com/en/7.x/parameters/#implementing-custom-types

src/hangar/cli/cli.py
rlizzo: Thanks for the feedback! I prefer the prefix `int:5` as well. I'm still hoping there might be something we can use which is a bit "shorter" than prepenting every integer with `int:`, but we may be out of options.

setup.py
hhsecond: But it's an extra_require and will be installed only if user do `pip install hangar[all]` or `pip install hangar[plugins]`. And we needed to give hangar as **batteries included** anyways. What do you think?

src/hangar/cli/cli.py
hhsecond: Thx

src/hangar/external/plugin_manager.py
```diff
@@ -0,0 +1,100 @@
...
+        """Load installed plugin.
+
+        User either needs to specify which plugin to load or should provide
+        file format to infer which plugin to use
+
+        Parameters
+        ----------
+        method : str
+            Which method to import from the plugin. Methods implemented by the
+            extension author should be publically announced in ``valid_provides``
```
hhsecond: @rlizzo There is a slight statement. `valid_provides` is the list of `methods` `PluginManager` says valid. It's not from plugin developers. Or am I missing something here?

src/hangar/cli/cli.py
hhsecond: The existing logic had changed here. Now instead of our CLI creating the file name, we let the plugin create the file names (this is because we are allowing writing to the same file in case of CSV, JSON etc). So I would assume plugin creators should handle such cases. If you think it makes sense, we could pass the type also along with sample name for the plugin to handle such cases but I would refrain from making another argument for the plugin function since there are workarounds. For answering your concerns
- I see the problem here but I would say, that does not seem like a frequent use-case (having sample name with both int and str of the same instance) and we could ignore it for now since I can't think of a sensible fix?
- The plugin could handle it
- So essentially, I skipped writing tests for it since it's a plugin thing

src/hangar/cli/cli.py
hhsecond: Done, thanks for pointing it out

setup.py
rlizzo: I disagree. The ability to do a `pip install[foo]` is as hidden a feature as any in the crazy world of python packaging. I didn't even know it was a thing until fairly recently. If we're going to write documentation on how to install and use these plugins, it's just easier to write the same "how-to" documentation and say to just treat it like any normal python plugin package:

    $ pip install hangar_matplotlib hangar_pil

More importantly than anything, if we include these as an option in the setup script, we need to test them fully. Any `pip install hangar` should be fully covered by our testing and performance guarrentees. We wouldn't be treating them as a "plugin", but as something "we plug in to". I agree that we should be "batteries included", but we are "batteries included" for the things which are relevant to Hangar: efficiently storing, retrieving, accessing, sharing, and collaborating on huge sets numerical data. The reason this project even works is because we don't care an ounce about the domain specific meanings that some set of stored bytes has. We just care about how we can most effectively get it in and out. 

I don't want to have to care about new releases of `matplotlib` or `pil` when I go to push a release, it's irrelevant to us, and a risk to include them as even something which could remotely be construed as "something the core team guarantees/supports".

src/hangar/cli/cli.py
hhsecond: Nice stuff!

src/hangar/cli/cli.py
rlizzo: > The existing logic had changed here. Now instead of our CLI creating the file name, we let the plugin create the file names (this is because we are allowing writing to the same file in case of CSV, JSON etc). So I would assume plugin creators should handle such cases.

Ok. This makes sense to me. 

> * I see the problem here but I would say, that does not seem like a frequent use-case (having sample name with both int and str of the same instance) and we could ignore it for now since I can't think of a sensible fix and since the plugin could handle it?
> * The plugin could handle it
> * So essentially, I skipped writing tests for it since it's a plugin thing

It doesn't matter if it's a frequent use case or not, it's something that we support (for better or worse in this instance) and since we won't control plugin code quality, we have to assume people will use this in strange ways with bad or just quickly written code. Sometime in the future when a piece of data overwrites itself on disk, the user is going to come and yell at us about the problem instead of going to the plugin developer. And honestly, i wouldn't really blame them, since they are using the `hangar` cli command after all. It's a minor thing now, and a lot of work to fix (clearly), but at the end of the day _every line of code has some degree of risk associated with it_, as library maintainers it's our job to minimize risk for our users. That's why they would use this over some home-grown solution at all.

As a side note, if you see a case where tests wouldn't pass because we don't have a piece of logic in place, it's better to write them and use `@pytest.mark.xfail(reason='foo', strict=True)` so we don't forget about smaller things like this it in the future. 

> So I would assume plugin creators should handle such cases.
> If you think it makes sense, we could pass the type also along with sample name for the plugin to handle such cases but I would refrain from making another argument for the plugin function since there are workarounds.

I agree this is a case where plugins creators handle this issue. It's just a matter of making a decision on how we should present this information to them. Passing a type along with the name isn't really any different from just passing a name as an `int` or `str`. If it's ignored on the plugin side, formatting the file name will just transparently cast and `int -> str`. 

We can eliminate this issue almost entirely is we do something like this:

1. we keep the same user facing selection API as discussed above 

```
    int:123   int
    str:123   str
    234       str
    foo1      str
    foo       str
    str:bar   str
```

But instead of passing a something we call `name` which would be either type `int` or `str`, every single sample name the plugin receives would be a string with the first 4 chars being `"type" + ":"`.

 It is up to that plugin to explicitly deal with how they think they should handle the names (in a way which is most convenient for that domain's users). If we just send them data like that, we are kindof forcing the plugin developers to deal with that decision and not get lazy. Even if they do, the user would have a file name with the type explicitly declared in it. 

```
    foo     -> 'str:foo'
    bar     -> 'str:bar'
    int:123 -> 'int:123'
    str:123 -> 'str:123'
    str:baz -> 'str:baz'
```

We document this behavior, and let the plugin authors deal with stripping the characters and formatting the name (or not, and it's OK for the user at the end of the day). 

Does this work?

src/hangar/external/plugin_manager.py
rlizzo: Ahh my mistake. So if I'm looking at this correctly now, this should read somehting like 

```
Which method to import from the plugin. Methods implemented by the
extension author should be declared as arguments passed into the 
BasePlugin superclass constructor at the time a plugin subclass is loaded
by the hangar plugin manager?
```

(probably in simpler terms?)

src/hangar/cli/cli.py
hhsecond: Thanks Rick, This is perfect

PR #153, opened 2019-10-28
BUG FIX: Zero Valued Arrays Hash Incorrectly

rlizzo:

Upon adding a small set of property based tests, it was discovered that zero valued arrays in variable shaped arraysets could potentially hash to the same value if two arrays had the same `size` and/or `itemsize` but different `shapes`

```python
>>> aset = co.arraysets.init_arrayset('aset', shape=(10, 10, 10, 10), dtype=np.float32, variable_shape=True)

>>> a = np.zeros(shape=(1, 1, 5, 1), dtype=np.float32)
>>> a.shape
(1, 1, 5, 1)
>>> b = np.zeros(shape=(1, 1, 1, 5), dtype=np.float32)
>>> b.shape
(1, 1, 1, 5)

>>> aset['a'] = a
>>> aset['b'] = b

>>> aset['a'].shape
(1, 1, 5, 1)
>>> aset['b'].shape
(1, 1, 5, 1)  # THIS IS INCORRECT!
```

This happens because 

```python
>>> a = np.zeros(shape=(1, 1, 5, 1), dtype=np.float32)
>>> b = np.zeros(shape=(1, 1, 1, 5), dtype=np.float32)
>>> a.tobytes() == b.tobytes()
True

>>> a_float = np.zeros(shape=(1, 1, 5, 1), dtype=np.float32)
>>> b_int = np.zeros(shape=(1, 1, 5, 1), dtype=np.int32)
>>> a_float.tobytes() == b_int.tobytes()
True
```

To fix this we now take the hash of the array bytes, with a binary struct composed of the ints making up each dimension shape and the numeric typecode of the array's dtype. 

```python
hasher = hashlib.blake2b(data, digest_size=20)
hasher.update(struct.pack(f'<{len(data.shape)}QB', *data.shape, data.dtype.num))
full_hash = hasher.hexdigest()
```

This means that any difference in type or shape (even if bytes are technically identical) will result in storage of that new array on disk and (more importantly) storage of the digest record (with corresponding variance in result spec) in the `hashenv` db. 

In addition to including the hypothesis test which found this in the test suite, manual tests have been added which assert this result does not happen. 

**This is a BREAKING CHANGE for all future versions of hangar, data hashs will no longer be the same** However, this is OK since that was going to happen in 0.4 anyways. 

## Types of changes
- [ ] Documentation update

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@ pip-log.txt
 nosetests.xml
 coverage.xml
 htmlcov
+.hypothesis
 
 # Translations
 *.mo
--- a/src/hangar/arrayset.py
+++ b/src/hangar/arrayset.py
@@ -9,6 +9,7 @@ import json
 
 import lmdb
 import numpy as np
+import struct
 
 from .backends import BACKEND_ACCESSOR_MAP
 from .backends import backend_decoder
@@ -765,7 +766,9 @@ class ArraysetDataWriter(ArraysetDataReader):
             if tmpconman:
                 self.__enter__()
 
-            full_hash = hashlib.blake2b(data.tobytes(), digest_size=20).hexdigest()
+            hasher = hashlib.blake2b(data, digest_size=20)
+            hasher.update(struct.pack(f'<{len(data.shape)}Q', *data.shape))
+            full_hash = hasher.hexdigest()
             hashKey = hash_data_db_key_from_raw_key(full_hash)
 
             # check if data record already exists with given key
--- a/src/hangar/arrayset.py
+++ b/src/hangar/arrayset.py
@@ -767,7 +767,7 @@ class ArraysetDataWriter(ArraysetDataReader):
                 self.__enter__()
 
             hasher = hashlib.blake2b(data, digest_size=20)
-            hasher.update(struct.pack(f'<{len(data.shape)}Q', *data.shape))
+            hasher.update(struct.pack(f'<{len(data.shape)}QB', *data.shape, data.dtype.num))
             full_hash = hasher.hexdigest()
             hashKey = hash_data_db_key_from_raw_key(full_hash)
 
--- a/src/hangar/remote/client.py
+++ b/src/hangar/remote/client.py
@@ -353,7 +353,9 @@ class HangarClient(object):
         for record in unpacked_records:
             data = chunks.deserialize_record(record)
             hasher = hashlib.blake2b(data.array, digest_size=20)
-            hasher.update(struct.pack(f'<{len(data.shape)}Q', *data.shape))
+            hasher.update(
+                struct.pack(f'<{len(data.array.shape)}QB', *data.array.shape,
+                            data.array.dtype.num))
             received_hash = hasher.hexdigest()
             if received_hash != data.digest:
                 raise RuntimeError(f'MANGLED! got: {received_hash} != requested: {data.digest}')
--- a/src/hangar/remote/client.py
+++ b/src/hangar/remote/client.py
@@ -4,6 +4,7 @@ import os
 import tempfile
 import time
 from typing import Tuple, Sequence
+import struct
 
 import blosc
 import grpc
@@ -351,7 +352,9 @@ class HangarClient(object):
         unpacked_records = chunks.deserialize_record_pack(uncompBytes)
         for record in unpacked_records:
             data = chunks.deserialize_record(record)
-            received_hash = hashlib.blake2b(data.array.tobytes(), digest_size=20).hexdigest()
+            hasher = hashlib.blake2b(data.array, digest_size=20)
+            hasher.update(struct.pack(f'<{len(data.shape)}Q', *data.shape))
+            received_hash = hasher.hexdigest()
             if received_hash != data.digest:
                 raise RuntimeError(f'MANGLED! got: {received_hash} != requested: {data.digest}')
             received_data.append((received_hash, data.array))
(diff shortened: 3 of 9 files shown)


rlizzo:
@hhsecond please review asap.

rlizzo:
Thanks for the quick review @lantiga. I'll hold off on merging this until we get a chance to talk this over in our weekly meeting. 

> Another solution would be to content-address using the hash computed from the bytes only, but verify that something has changed through the newer hash. This would save us from storing again after a change in shape,

I think theres some confusion on one of our ends, because this doesn't make much sense to me. Not sure if i'm missing your point, or if something about the actual implementation is getting lost. 

> It would be good to know how performance is impacted by this change - very likely negligibly.

The only time this could possibly affect performance are in two sections:

1) adding data -> when the full (cryptographic) digest of an input sample is actually calculated and stored. 
2) transferring data between a client/server -> The cryptographic hash is calculated (once again) so that the receiver assures the transfered data is what the sender said it was going to send (god, that's a horrible way to phrase that sentence, sorry!) 

When reading data, we don't recompute the cryptographic digest on every read (that would be far too inefficient). We use [xxHash](https://github.com/Cyan4973/xxHash) as an extremely fast, well distributed, vetted, but non-cryptographic method to validate that `bytes out == bytes expected`. 

I'll run the benchmark suite (still need to propose that as a PR in master) but I'm expecting nearly no measurable change whatsoever. packing a tuple of python ints which is _at most 33 elements long_ (the maximum number of dimensions we handle, plus the `dtype num`) - and which in the vast majority of cases will be less than 5 items long - into a `C-struct` is extremely efficient.  

Your continual reminders about performance patterns and memory consumption has not gone un-noticed. It's seriously influenced the ways I develop for Hangar, and we're all better because of it! Thanks for brining up the concern though, I'll post the benchmarks when I have them run here.

rlizzo:
> * adding data -> when the full (cryptographic) digest of an input sample is actually calculated and stored.
> * transferring data between a client/server -> The cryptographic hash is calculated (once again) so that the receiver assures the transfered data is what the sender said it was going to send (god, that's a horrible way to phrase that sentence, sorry!)

I never explained why these are areas where even if we take a hit, it's inconsequential... 

For point (1), I've spent a serious amount of time profiling and tracing execution along this codepath. The vast majority of time spent in the add data method (well over 95%) is in compressing data and flushing it to disk. The other 5% is nearly completly spent calculating the digest of the `array bytes`. As arrays increase in size, this can go to something more like a 90%-10% split, but the execution is still dominated by the backend storage IO/compression. Since the `hasher.update()` call is an incremental update to the digest, and we are passing in very small amounts of data, this should be nearly an unmeasurable addition. 

For point (2), the same logic as in point(1) applies (since you still need to save the data to disk after calculating the digest, but execution is **absolutely dominated by network transfer speeds**. Once the data is transferred to/from the client/server, the amount of time spent writing to disk is laughable compared to how long it took to transfer over a network.

rlizzo:
And to asuade your fears, here is proof from the actual benchmark suite we have:

```
[:~ … hangar-benchmarks-local] [hangar-dev] 1 $ asv compare master property-based-testing

All benchmarks:

       before           after         ratio
     [8d9515bc]       [a0a14620]
     <master>         <property-based-testing>
            57.7M            57.5M     1.00  arrayset_sample_reads.HDF5_00.peakmem_read_float32_2000_samples
            55.6M            55.5M     1.00  arrayset_sample_reads.HDF5_00.peakmem_read_uint16_2000_samples
          2.09±0s          2.15±0s     1.03  arrayset_sample_reads.HDF5_00.time_read_float32_2000_samples
          1.53±0s          1.52±0s     1.00  arrayset_sample_reads.HDF5_00.time_read_uint16_2000_samples
            1.55G            1.55G     1.00  arrayset_sample_reads.NUMPY_10.peakmem_read_float32_2000_samples
             803M             803M     1.00  arrayset_sample_reads.NUMPY_10.peakmem_read_uint16_2000_samples
          926±0ms          926±0ms     1.00  arrayset_sample_reads.NUMPY_10.time_read_float32_2000_samples
          474±0ms          481±0ms     1.02  arrayset_sample_reads.NUMPY_10.time_read_uint16_2000_samples
          10.7±0s          10.5±0s     0.98  arrayset_sample_writes.HDF5_00.time_add_float32_samples(2000)
          6.26±0s          6.11±0s     0.98  arrayset_sample_writes.HDF5_00.time_add_uint16_samples(2000)
             165M             165M     1.00  arrayset_sample_writes.HDF5_00.track_repo_size_float32_samples(2000)
             111M             111M     1.00  arrayset_sample_writes.HDF5_00.track_repo_size_uint16_samples(2000)
          4.03±0s          4.06±0s     1.01  arrayset_sample_writes.NUMPY_10.time_add_float32_samples(2000)
          2.10±0s          2.08±0s     0.99  arrayset_sample_writes.NUMPY_10.time_add_uint16_samples(2000)
             1.5G             1.5G     1.00  arrayset_sample_writes.NUMPY_10.track_repo_size_float32_samples(2000)
             751M             751M     1.00  arrayset_sample_writes.NUMPY_10.track_repo_size_uint16_samples(2000)
         366±10ms          370±6ms     1.01  package.TimeImport.time_import
```

Any observed changes are well within a margin of error between test runs.

PR #159, opened 2019-11-07
BugFix: Removing All Samples in Arrayset Should Not Remove Aset Schema

rlizzo:

In previous versions of hangar, when the last sample in an arrayset was deleted, the arrayset schema would implicitly also be deleted (with no user notice). 

```python
>>> len(co.arraysets)
1
>>> len(co.arraysets['foo'])
0
>>> co.arraysets['foo'][1] = np.arange(10)
>>> len(co.arraysets['foo'])
1
>>> del co.arraysets['foo'][1]
>>> len(co.arraysets['foo'])
traceback
---------------------------------------
ReferenceError: 'foo' no longer exists
...
>>> len(co.arraysets)
0
```

In addition to being a detriment to UX, the level at which this operation occurred (`ArraysetDataWriter`) meant that the change was not being propagated up to the `Arraysets` class which holds the only strong reference keeping the object alive. Since the strong reference was never deleted, the user could technically still hold a live weakref proxy to this object which should have been finalized. 

In the worst case scenario (if a context manager for _any_ `Arrayset` class was open at the time the last sample was removed), backend file handles may not be closed and invalidated properly (which would force an exception on any `set` or `get` operation). Should this happen, it was possible that a `set` to the `ArraysetDataWriter` would actually succeed - saving data to disk and writing a (valid) record reference to the staging db. Though the sample was recorded, the record of the schema spec would have been removed from the staging area / commit refs as soon as the `Arrayset` implicitly deleted itself. Upon a later checkout of this (or a child) commit/branch, the schema spec corresponding to the added sample refs would not exist, and as such no `Arrayset` would be generated for the sample (even though a valid reference was present, and a valid schema spec may exist in the commit's ancestory)

```python
>>> co.arraysets['foo'][1] = np.arange(10)
>>> len(co.arraysets['foo'])
1
>>> aset = co.arraysets['foo']
>>> with co:
...    del co.arraysets['foo'][1]
>>> len(co.arraysets)
1
>>> with aset:
...    aset[2] = np.zeros(10)
>>> len(aset)
1
>>>
>>>
>>> co.commit('should not work')
foohash
>>> co.close()
>>> co = repo.checkout(write=True)
>>> len(co.arraysets)
0
>>> co.arraysets['foo']
traceback
--------------------------
KeyError: 'foo' does not exist
```

- Removing all samples from an arrayset no longer deletes the arrayset spec
- Initializing or removing an `Arrayset` can not be performed when any arrayset is open in a context manager. 
- calculation of schema hash digest (and all hash digest functions) made more deterministic and moved into isolated module. 
- tests added ensuring correct behavior

## Types of changes
- [ ] Documentation update

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:


src/hangar/arrayset.py
```diff
@@ -1093,7 +1084,8 @@ def items(self) -> Iterable[Tuple[str, Union[ArraysetDataReader, ArraysetDataWri
         Iterable[Tuple[str, Union[:class:`.ArraysetDataReader`, :class:`.ArraysetDataWriter`]]]
             returns two tuple of all all arrayset names/object pairs in the checkout.
         """
-        for asetN, asetObj in self._arraysets.items():
+        for asetN in list(self._arraysets.keys()):
```
hhsecond: just curious: why an explicit `list()` here?

src/hangar/arrayset.py
```diff
@@ -1278,6 +1282,8 @@ def init_arrayset(self,
 
         Raises
         ------
+        PermissionError
+            If any enclosed arrayset is opned in a connection manager.
```
hhsecond: opened

src/hangar/arrayset.py
rlizzo: Yeah.. I hate it too... It's because of the following situation 

```python
>>> for name, aset in co.arraysets.items():
>>>     del co.arraysets[name]

traceback RuntimeError
------------------------
RuntimeError: dictionary changed size during iteration
```
While I normally subscribe to the belief that "if you mutate a data structure while iterating over it, you are performing a mortal sin and deserve whatever happens to you", it's hangars responsibility to manager  usage of  this thing which kindof behaves like a `dict` and class simultaneously. Seemed unfair to put the implications of dealing with an implementation detail on the user when it can be fixed so trivially...

PR #173, opened 2019-12-04
Validate Repo Integrity

rlizzo:

Validate the integrity of repo data and commit history.

- crytographic hash of all data (read from disk) and commit refs
- validate integrity of commits, data, branches, and historical provenance. 
- working in manual tests, not automated yet. 

## Types of changes
- [ ] Documentation update

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:


rlizzo:
@hhsecond can you take a look at this. The tests aren't complete, it needs documentation, and there are a few corner cases to take care of, but I'd like your opinion on if I'm missing anything significant. 

My TODO list right now is:

- Handle cases where key / values are slightly malformed. We need to catch exceptions thrown by any parsing methods encountered on the way to the `integrity` module (I'll highlight concerning regions below)
- Acquire the writer lock before running. These methods don't actually write anything, but it wouldn't be fun to deal with changing world state while the checks run.

**Also, i'm not sure if I want to be raising a `RuntimeError` at every point some issue is found?** I like the current implementation because:

1. We don't do any unnecessary work. As soon as an issue is encountered the world comes crashing down and everything stops. 
2. Deliminating issues in the code via a prominent `raise RuntimeError(msg)` statement is rather nice in terms of readability. There's no guessing what the issue encountered is, or where it was found since theres a detailed error message right there.

**HOWEVER, raising errors in this fashion means that it's not really possible to generate a list of issues should there be multiple things wrong .** In addition to stopping at the first error it encounters, this doesn't return a machine readable description of the issue. In its current form, there's not any reasonable path towards effective logging or automated attempts to recover data.  

These seem like issues for the future, but I'm not sure if you have other thoughts?

src/hangar/diagnostics/integrity.py
```diff
@@ -0,0 +1,212 @@
...
+        tcode = hashmachine.hash_type_code_from_digest(digest)
+        calc_digest = hashmachine.schema_hash_digest(
+            shape=val.schema_max_shape,
+            size=np.prod(val.schema_max_shape),
+            dtype_num=val.schema_dtype,
+            named_samples=val.schema_is_named,
+            variable_shape=val.schema_is_var,
+            backend_code=val.schema_default_backend,
+            backend_opts=val.schema_default_backend_opts,
+            tcode=tcode)
```
rlizzo: This whole section is particularly susceptible to breakage. If the schema value is corrupted or changed, theres no guarantee that the parsed values would contain the expected field names (if it was even able to be deserialized at all)

src/hangar/diagnostics/integrity.py
```diff
@@ -0,0 +1,212 @@
...
+            if val.backend not in fs:
+                fs[val.backend] = BACKEND_ACCESSOR_MAP[val.backend](repo_path, None, None)
+                fs[val.backend].open(mode='r')
+            if is_local_backend(val) is False:
+                nremote += 1
+                warnings.warn(
+                    'Can not verify integrity of partially fetched array data references. '
+                    'For complete proof, fetch all remote data locally.', RuntimeWarning)
+                continue
+            o = fs[val.backend].read_data(val)
```
rlizzo: Same issue exists here as for `_verify_schema_integrity`, there's no guarrentee that the values would even be parsable, let alone contain the expected field names. 

Additionally, these variable names suck. How was I possibly ok with them wile writing this??

rlizzo:
TODOS complete and tests finished. When you have a chance, would love to hear your thoughts @hhsecond. No rush.

src/hangar/records/hashs.py
```diff
@@ -1,9 +1,16 @@
...
+from .parsing import (
+    arrayset_record_schema_raw_key_from_db_key,
+    arrayset_record_schema_raw_val_from_db_val,
+    hash_data_raw_key_from_db_key,
+    hash_meta_raw_val_from_db_val,
+    hash_schema_raw_key_from_db_key,
+)
 
 
 class HashQuery(object):
```
hhsecond: A short docstring for this class could help

src/hangar/records/hashs.py
```diff
@@ -13,68 +20,178 @@ def __init__(self, hashenv: lmdb.Environment):
...
-                while hashsExist:
-                    hashRecKey, hashRecVal = cursor.item()
-                    if hashRecKey.startswith(startHashRangeKey):
-                        yield hashRecKey
-                        hashsExist = cursor.next()
-                        continue
-                    else:
-                        hashsExist = False
+                # divide loop into returned type sections as perf optimization
+                # (rather then if/else checking on every iteration of loop)
```
hhsecond: out of curiosity: Was this performance gain significant?

src/hangar/records/hashs.py
```diff
@@ -13,68 +20,178 @@ def __init__(self, hashenv: lmdb.Environment):
...
+                else:  # pragma: no cover
+                    raise ValueError(f'Both keys and values argument cannot be False')
         finally:
             TxnRegister().abort_reader_txn(self._hashenv)
 
     def list_all_hash_keys_raw(self):
-        recs = self._traverse_all_hash_records()
-        out = list(map(parsing.hash_data_raw_key_from_db_key, recs))
+        recs = self._traverse_all_hash_records(keys=True, values=False)
+        out = list(map(hash_data_raw_key_from_db_key, recs))
```
hhsecond: list comprehension? Here and few other places where we have map -> list conversion

src/hangar/dataloaders/tfloader.py
```diff
@@ -1,12 +1,17 @@
...
 
 from .common import GroupedAsets
+from ..utils import LazyLoader
 
 try:
-    import tensorflow as tf
+    tfExists = importlib.util.find_spec('tensorflow')
+    if not tfExists:
+        raise ModuleNotFoundError
+    tf = LazyLoader('tf', globals(), 'tensorflow')
```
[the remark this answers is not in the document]
rlizzo: Thanks. No idea... Wasn't thinking about the effect on import time apparently.

src/hangar/records/hashs.py
rlizzo: Well technically there is no direct performance gain over the "old" method, since it didn't do the checks, but check the following benchmark for the actual saves we do see:

```python
iterlist = [(f'k_{i}', f'v_{i}') for i in range(100_000)]

def run_slow(keys: bool = True, values: bool = True):
    for db_k, db_v in iterlist:
        if (keys is True) and (values is True):
            yield (db_k, db_v)
        elif (keys is True) and (values is False):
            yield db_k
        elif (keys is False) and (values is True):
            yield db_v
        else:
            raise ValueError('both keys and values param cannot be false')
            
def run_fast(keys: bool = True, values: bool = True):
    if (keys is True) and (values is True):
        for db_k, db_v in iterlist:
            yield (db_k, db_v)
    elif (keys is True) and (values is False):
        for db_k, _ in iterlist:
            yield db_k
    elif (keys is False) and (values is True):
        for _, db_v in iterlist:
            yield db_v
    else:
        raise ValueError('both keys and values param cannot be false')
```
```python
%timeit list(run_slow(keys=True, values=True))
%timeit list(run_slow(keys=True, values=False))
%timeit list(run_slow(keys=False, values=True))
```

```
12.6 ms ± 374 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
10.9 ms ± 289 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
10.9 ms ± 204 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

```python
%timeit list(run_fast(keys=True, values=True))
%timeit list(run_fast(keys=True, values=False))
%timeit list(run_fast(keys=False, values=True))
```

```
10.4 ms ± 110 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
6.72 ms ± 193 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
6.76 ms ± 311 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

src/hangar/records/hashs.py
rlizzo: list comprehension is ~10% slower than mapping when the function is predefined. I've changed it where appropriate, but won't be making the change universally.

hhsecond: But since we have a conversion from map object to `list`, I don't think the performance gain map could bring is relevant anymore. It's going to be slower than list comprehension, IMO. I mean, it's probably negligible but just saying if you want to do a comparison

src/hangar/records/hashs.py
```diff
@@ -1,80 +1,213 @@
...
+from ..txnctx import TxnRegister
 
 
 class HashQuery(object):
+    """Traverse and query contents contained in ``hashenv`` and ``labelenv`` dbs
+
+    These methods operate on the databases which store the mapping of some data
+    digest to it's location on disk (or value in the case of metadata and
+    schemas). These databases are not specific to a particular commit; the
+    records are are for every piece of data data stored in every commit across
```
hhsecond: are are -> are
data data -> data

PR #179, opened 2020-01-04
Arrayset Subsamples

rlizzo:

This is a large PR which started with the motivation of allowing arraysets to contain subsamples under a common key. Though minimal work was needed for the technical implementation (with esentially no changes made to the hangar core record parsing, history traversal, or tensor storage backends), the integration of the API into the current model proved difficult, which required some major refactoring of what was previously known as `ArraysetDataReader` and `ArraysetDataWriter` classes. 

Rather than try to combine every possible API method needed by `flat` and `nested` arrayset access into a frankenstein monster class, each access convention implements it's own API class methods (fully independent from one another). The appropriate constructors are selected based on the `constains_subsamples` argument in `init_arrayset()`. The argument is recorded in the schema so the correct type can be identified in subsequent checkouts.

I'm working on putting together a summary of the API. That will follow shortly.

At the moment, about half the tests for the new `nested` sample container are missing, and I need to re-evaluate some implementation details for how backend file handles are dealt with. 

## Types of changes
- [ ] Documentation update

- [ ] Ready for review
- [x] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:


src/hangar/checkout.py
```diff
@@ -64,15 +60,15 @@ class ReaderCheckout(object):
     """
 
     def __init__(self,
-                 base_path: os.PathLike, labelenv: lmdb.Environment,
+                 base_path: Path, labelenv: lmdb.Environment,
```
hhsecond: it could be both `Path` and `str`?

src/hangar/cli/cli.py
```diff
@@ -175,15 +176,22 @@ def create_arrayset(repo: Repository, name, dtype, shape, variable_, named):
...
         co = repo.checkout(write=True)
         aset = co.arraysets.init_arrayset(name=name,
                                           shape=shape,
                                           dtype=np.typeDict[dtype.lower()],
-                                          named_samples=named,
-                                          variable_shape=variable_)
-        click.echo(f'Initialized Arrayset: {aset.name}')
+                                          variable_shape=variable_,
+                                          contains_subsamples=subsamples_)
+        click.echo(f'Initialized Arrayset: {aset.arrayset}')
```
hhsecond: IMO `name` was more intuitive than `arrayset`

setup.py
```diff
@@ -1,23 +1,199 @@
...
+import platform
+import sys
+from distutils.sysconfig import get_config_var
+from distutils.version import LooseVersion
 from glob import glob
-from os.path import basename
-from os.path import dirname
-from os.path import join
-from os.path import splitext
+from os.path import basename, join, splitext
```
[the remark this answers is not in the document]
rlizzo: `distutils` does not accept `Path`-like objects as the source for compiled c extensions (needed for cython)

src/hangar/checkout.py
rlizzo: no, this is internal to hangar. We expect a `Path` object.

src/hangar/cli/cli.py
rlizzo: yeah, but when we update the internal naming from `arrayset ->columns`, this is going to have to be more descriptive than `name`.  I'm actually going to break this `NamedTuple` class into `column-type` specific `NamedTuple` classes. It's going to be messy to refactor, this will just help in the future (it's a temporary name).

PR #180, opened 2020-02-19
Column API and DataType Containers

rlizzo:

This PR implements the new `columns` API, replacing the old `arraysets` and `metadata` terminology. The concept of a column is to act as a highly modular container for potentially arbitrary data types. As a simple proof of concept, a `string` type column has been introduced - replacing the functionality of `metadata` in a much more useful and expandable way.

More details will be posted soon. 

## Types of changes
- [ ] Documentation update

- [ ] Ready for review
- [x] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:


src/hangar/typesystem/descriptors.py
```diff
@@ -0,0 +1,187 @@
...
+    pass
+
+
+@MaxSized
+@TypedSequence(int)
+class SizedIntegerTuple(Tuple):
+    pass
+
+
+class checkedmeta(type):
```
hhsecond: `__set_name__` does exactly this. We don't need this metaclass perhaps.

src/hangar/backends/__init__.py
```diff
@@ -1,30 +1,146 @@
...
+process.
+"""
+import string
+from typing import Union, Dict
 
 from .specs import (
     HDF5_00_DataHashSpec,
     HDF5_01_DataHashSpec,
     NUMPY_10_DataHashSpec,
+    LMDB_30_DataHashSpec,
```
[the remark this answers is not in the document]
rlizzo: I already have one of the hangar Enterprise backends already assigned to `20`. It might make more sense for the different projects to take backend codes starting on different ends of the available key space tho. Good call.

src/hangar/typesystem/descriptors.py
rlizzo: I'll take a look, thanks for pointing that out!

rlizzo:
@hhsecond any idea why this test is failing? I'm not sure what tensorflow wants here...

https://travis-ci.com/tensorwerk/hangar-py/jobs/289376731#L423-L506

src/hangar/dataloaders/common.py
```diff
@@ -113,14 +113,14 @@ def get_shapes(self, converter=None):
...
         -------
-        A tuple of arrayset shapes
+        A tuple of column shapes
         """
         shapes = []
-        for aset in self.arrayset_array:
-            if aset.variable_shape:
-                aset_shape = (None,) * len(aset.shape)
+        for col in self.columns_col:
+            if col.column_layout == 'variable_shape':
```
hhsecond: This should be `col.schema_type`

hhsecond:
> @hhsecond any idea why this test is failing? I'm not sure what tensorflow wants here...
> 
> https://travis-ci.com/tensorwerk/hangar-py/jobs/289376731#L423-L506

@rlizzo [Here](https://github.com/tensorwerk/hangar-py/blob/3282d4919df02d56df7de6f72ed38c8cd8fdabac/src/hangar/dataloaders/common.py#L120) is the culprit

src/hangar/columns/column.py
```diff
@@ -0,0 +1,701 @@
...
+                    if k.startswith(startRangeKey):
+                        recordsExist = cursor.delete()
+                    else:
+                        recordsExist = False
+            datatxn.delete(columnSchemaKey)
+
+        return column
+
+    @writer_checkout_only
+    def create_str_column(self,
```
hhsecond: @rlizzo @lantiga 
IMO; these names are verbose and maybe we could try something else. Few ideas on top of my mind (which needs polishing) are
- `co.columns.create(type='ndarray' / 'str', layout='flat' / 'nested')` where the argument names are arbitrary and could be anything else
- Constructors being nouns like `np.array` -> `co.column.flat`, `co.column.nested`, `col.column.str`

src/hangar/backends/__init__.py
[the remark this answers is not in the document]
elistevens: Maybe enterprise backends should be three digit?

src/hangar/columns/column.py
elistevens: I think that verbosity is a feature here; customers need that verbosity when they're trying to learn the API.

I don't think that `co.columns.create` is particularly Pythonic, since the intermediate `columns` object is tied to and operating on the `co`, but presumably is also a dict-like thing. Calling create on a dict to manipulate the dict's parent object? That's weird.  Also, `flat` has no meaning to me, though perhaps that will be fixed once the v0.5.0 docs are live.

src/hangar/dataloaders/common.py
rlizzo: Ugh. Thanks!

Such a stupid mistake to make... I spent way more time trying to get that test to work then you would like to know...

src/hangar/columns/column.py
hhsecond: When I said verbose, I meant, we are repeating information here. The API in this PR is `co.columns.create_xxx_column` and by this, we are repeating the word `column` which would be obvious even if we use `co.columns.create()`.  I am not particularly sentimental about the name `create`. It could be anything else, like `add` or `update` (like `dict.update`, thinking aloud here).

elistevens: Ah, I misunderstood, sorry.

I think the function should go on the checkout object; putting it on the columns pseudo-container seems odd.

docs/Tutorial-Dataloader.ipynb
```diff
@@ -105,7 +105,7 @@
     "repo.init(user_name=username, user_email=email, remove_old=True)\n",
     "co = repo.checkout(write=True)\n",
     "\n",
-    "images_aset = co.arraysets.init_arrayset('images', shape=img_shape, dtype=np.uint8, named_samples=False)\n",
+    "images_aset = co.columns.init_arrayset('images', shape=img_shape, dtype=np.uint8, named_samples=False)\n",
```
hhsecond: It's probably "WIP". Just making  sure that you didn't miss this one out

docs/concepts.rst
```diff
@@ -91,7 +91,7 @@ time by various "Committers" across any number of "Branches". Though there are
 many conceptual similarities in what a Git repo and a Hangar Repository
 achieve, Hangar is designed with the express purpose of dealing with numeric
 data. As such, when you read/write to/from a Repository, the main way of
-interaction with information will be through (an arbitrary number of) Arraysets
+interaction with information will be through (an arbitrary number of) Columns
 in each Commit. A simple key/value store is also included to store metadata,
 but as it is a minor point is will largely be ignored for the rest of this
```
hhsecond: May be mention that this metadata is commit level and will not be part of the history.

docs/concepts.rst
```diff
@@ -474,21 +474,21 @@ operations which can occur.
 
 :Addition:
 
-    An operation which creates a arrayset, sample, or some metadata which
+    An operation which creates a column, sample, or some metadata which
```
hhsecond: metadata will not be part of the history?

src/hangar/mixins/datasetget.py
```diff
@@ -314,7 +313,7 @@ def get(self, arraysets, samples, *, except_missing=False):
 
             # Data Retrieval
             asetsSamplesData = []
-            for aset in arraysets:
+            for aset in columns:
                 aset_samples = []
                 for sample in samples:
                     try:
```
hhsecond: `arr = aset.get(sample)` will not throw key error and return `default`. So even if the argument `except_missing` is `True`, `co['aset', 'wrongkey']` will not raise an error. I assume this line should be `aset[sample]`

hhsecond: Forgot to mention that this issue is not related to this PR. Something we have missed in the past review and our test suite couldn't catch

src/hangar/columns/column.py
rlizzo: Hey guys as a follow up for completeness here...

1) The method has been renamed to `add_ndarray_column()` and `add_string_column()`. This is to reflect the actual workings of the method. Me and @lantiga considered `define_foo_column()` as well (as suggested by @elistevens) but decided against it because while acting as a definition, the operation also creates a column ready for use. We wanted to make it clear that nothing else needs to be done in order to use the column that was "defined/initialized" 

2) The method does now live on the `checkout` class (instead of `checkout.columns`). This is just better API design. ex: 

```python
col = co.add_ndarray_column('foo', shape=(10, 10), dtype=np.uint8)
col[1] = np.arange(10, dtype=np.uint8)
co.commit('added column')
co.close()
```

3) I'm going to be working on improving docstrings and tutorials shortely, but with the new "generalized" column forms, the options essentially break down into the following:

- `"layout"`: the method by which data in columns are named/indexed/accessed by. This can currently take on the values of `"flat"` (for single-level story type mapping) OR `"nested"` (for a sample name which contains subsample name/data pairs)

- `"type"`: indicates the format of data stored within the column container. Valid values are `"ndarray"` or `"string"`. These are not accepted as keyword arguments, but are determined by the method name which is actually called to create the column (`add_ndarray_column()` for `"ndarray"` and `add_str_column()` for `"string"`)

- `"variable_shape"`: bool indicating if data length / size can be variable sized (will have different defaults and avalability depending on the column `"type"). For example, `ndarray` columns accept both `True` and `False` values, while `string` column types only accept `True`. 

- `"backend"` and `"backend_options"`: specify specific backend and any filter / compression options to apply to the column. These are advanced arguments which are type & value checked / enforced based on the definition of column values described above. I won't get into the full details here... 

Making this all clear is a high priority for me, and i'm going to need input on the explanations we eventually get around to writing. Let me know if you have any questions!

docs/Tutorial-Dataloader.ipynb
rlizzo: Got it!

docs/concepts.rst
rlizzo: This will be in a seperate PR.

src/hangar/backends/__init__.py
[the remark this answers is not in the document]
rlizzo: Not a bad idea, I'll look into it!

src/hangar/mixins/datasetget.py
rlizzo: Yup! Making the change now and adding a test!

docs/concepts.rst
rlizzo: Will be addressed in #181

PR #188, opened 2020-03-12
Hangar Real World Quick Start Tutorial

alessiamarcolini:
New tutorial covering only the basic stuff for version 0.5 release, such as Repository creation and initialization, adding data to columns and committing changes.

## Types of changes
- [x] Documentation update

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:


alessiamarcolini:
Thank you @lantiga for the review!

I agree with the first point, I will change the name of the checkout object to `master`.
I'll add a comment to clarify that would be impractical for larger datasets.

Also, why are some tests failing?

hhsecond:
@alessiamarcolini Tests are failing probably because of some version upgrade came to one of our test dependencies. Even the master is failing since last week. @rlizzo might have a clue

lantiga:
No idea actually, @rlizzo ?

elistevens:
Can I suggest `master_checkout` or `head_checkout`?  The term `master` by itself is also used for replication, and this helps make it clear that it's a `XxxCheckout` object.

When I've copied example code into other projects, reviewers have complained about it not being clear what `co` is.

Another idea: since there are many places in the API that only accept one or the other of reader or writer checkouts, perhaps the names should be `head_read_checkout` or `head_wcheckout` or something like that (I prefer the first, but it can get verbose).

lantiga:
Hey @elistevens :-) What about `master_branch`? Wouldn't it be less confusing than `master_checkout`? IDK, `checkout` sounds so SVN-y

elistevens:
If the object's class is `XxxCheckout` then I think that should be reflected in the name of the variable.

lantiga:
> If the object's class is `XxxCheckout` then I think that should be reflected in the name of the variable.

Agreed. It goes both ways is guess 😜

alessiamarcolini:
@rlizzo should be ready now :)

rlizzo:
Looks good! Merging!

PR #204, opened 2020-05-26
New Feature - Bulk Importer

rlizzo:

Bulk importer methods to ingest large quantities of data into Hangar.

The following module is designed to address challenges inherent to writing massive amounts of data to a hangar repository via the standard API. Since write-enabled checkouts are limited to processing in a single thread, the time required to import hundreds of Gigabytes (or Terabytes) of data into Hangar (from external sources) can become prohibitivly long. This module implements a multi-processed importer which reduces import time nearly linearly with the number of CPU cores allocated on a machine.

There are a number of challenges to overcome:

1. How to validate data against a column schema?

    - Does the column exist?
    - Are the key(s) valid?
    - Is the data a valid type/shape/precision valid for the the selected column schema?

2. How to handle duplicated data?

    -  If an identical piece of data is recorded in the repository already, only record the sample reference (do not write the data to disk again).
    - If the bulk import method would write identical pieces of data to the repository multiple times, and the data does not already exist, then that piece of content should only be written to disk once. Only sample references should be saved after that.

3. How to handle transactionality?

     - What happens if some column, sample keys, or data piece is invalid and cannot be written as desired?
    - How to rollback partial changes if the process is inturupted in the middle of a bulk import operation?

4. How to limit memory usage if many processes are trying to load and write large tensors?

Rough outline of steps:

    1. Validate UDF & Argument Signature
    2. Read, Validate, and Hash UDF results --> Task Recipe
    3. Prune Recipe
    4. Read, Validate, Write Data to Isolated Backend Storage
    5. Record Sample References in Isolated Environment
    6. If all successful, make isolated data known to repository core, otherwise abort to starting state.

## Types of changes
- [ ] Documentation update

- [x] Ready for review
- [ ] Work in progress

## How Has This Been Tested?
- [ ] I don't know. Help!

## Checklist:


rlizzo:
@elistevens you'll like this!

PR #216, opened 2020-09-02
Tests for new dataset APIs

hhsecond:
## Motivation and Context
Tests for new dataset APIs

### DIFF
--- a/tests/ml_datasets/test_dataset.py
+++ b/tests/ml_datasets/test_dataset.py
@@ -105,6 +105,13 @@ class TestInternalDatasetClass:
         co = repo_20_filled_subsamples.checkout()
         col1 = co['writtenaset']
         col2 = co['second_aset']
+
+        dataset = HangarDataset([col1, col2])
+        data = dataset.index_get(1)
+        assert tuple(data[0].keys()) == tuple(data[1].keys()) == (4, 5, 6)
+        assert isinstance(data[0], dict)
+        assert isinstance(data[1], dict)
+
         keys = (((0, ...), (0, 1)), ((1, ...), (1, 4)))
         dataset = HangarDataset([col1, col2], keys=keys)
         data = dataset.index_get(1)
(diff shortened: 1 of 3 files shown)


tests/ml_datasets/test_dataset.py
```diff
@@ -156,6 +163,47 @@ def test_multiple_dataset_batched_loader(self, repo_20_filled_samples):
...
+        # testing with batch_size = 1
+        dset = make_numpy_dataset([first_aset, second_aset], batch_size=1, drop_last=True)
+        total_samples = 0
+        for dset1, dset2 in dset:
+            total_samples += dset1.shape[0]
+            assert dset1.shape == (1, 5, 7)
+            assert dset2.shape == (1, 5, 7)
+        assert total_samples == 20  # drop last is True will not have any effect
+
+        with pytest.raises(RuntimeError):
```
rlizzo, line 176: Use `match='foo message'` in order to make sure the correct error is raised here.

tests/ml_datasets/test_dataset.py
```diff
@@ -156,6 +163,47 @@ def test_multiple_dataset_batched_loader(self, repo_20_filled_samples):
...
+            # Setting drop_last without batching
+            dset = make_numpy_dataset([first_aset, second_aset], batch_size=0, drop_last=True)
+        dset = make_numpy_dataset([first_aset, second_aset], batch_size=0)
+        total_samples = 0
+        for dset1, dset2 in dset:
+            total_samples += 1
+            assert dset1.shape == (5, 7)
+            assert dset2.shape == (5, 7)
+        assert total_samples == 20
```
rlizzo, line 185: Need a test where `drop_last=True` actually has an effect.

rlizzo (changes requested):
Two minor test additions which should be included. Otherwise looks good!

tests/ml_datasets/test_dataset.py
hhsecond: That was already there. On line #165

tests/ml_datasets/test_dataset.py
hhsecond: Done

